PR #3, opened 2021-11-04
Update issue templates

marcoep:
Old ones dont work for some reason..


marcoep:
@boris-vukovic the previous templates did not work... So, lets try again with these ones :D

PR #10, opened 2021-11-23
Github workflow for LabExT

maltewae:
# Goal
Setup of a CI for LabExT to run tests automatically for each push. This can also be used for linters later.

# Approach
Use of a Github action, which is triggered for every push. In addition, Tox is used, which is a wrapper for pytest. Tox first generates a package from the code and then calls pytest to run all tests (files with * _test.py).
Further information can be found at: https://github.com/tox-dev/tox

# Compatibility
This PR does not introduce any new functions in LabExT, but only sets up GitHub for CI. Nevertheless, all runs will fail because pytest has not yet found any tests and so exits with status code 1.

### DIFF
--- a/.github/workflows/labext-ci.yml
+++ b/.github/workflows/labext-ci.yml
@@ -1,0 +1,23 @@
+name: LabExT CI
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


marcoep (approved):
Great! However, so far we have used the unittest package so far. I assume you have seen this? I guess, we need to rewrite the tests in LabExT/Tests to make them pytest compatible?

maltewae:
Pytest will also run unittest test files. What we need to do is change the signature of the files to `_test.py`, which I can do in a follow up PR. Some tests (e.g https://github.com/LabExT/LabExT/blob/main/LabExT/Tests/Movement/Test_PiezoStage.py) will not run in CI because they need hardware. We have to omit these in the CI.
If it is ok for you, I would merge this PR and I will open up another PR to make the current tests compatible.

marcoep:
Absolutely, thanks!

marcoep:
@maltewae You can leave out adapting `Test_OpticalSpectrumAnalyzerAQ6370C.py` and `Test_PowerMeterGenericKeysight.py` as they require actual lab instruments to run...

PR #11, opened 2021-11-23
[Movement] Refactor: Introduce Stage interface and SmarAct implementation

maltewae:
# Goal
This PR introduces a new interface for all stages in LabExT, a stage implementation for SmarAct and unit tests.

# Approach
The idea is to have a unified interface for all stages in LabExT to make future devices easier to implement and have more consistency.
Important: To keep the changes in this PR small, the interface is the same as in [PiezoStage](https://github.com/LabExT/LabExT/blob/main/LabExT/Movement/PiezoStage.py), so existing code will continue to work. The interface can be revised later.
Mocks are used to enable testing. I propose to simulate the PythonWrapper module for SmarAct and assume that it works without errors. The written tests verify that LabExT calls the functions correctly and handles all return values correctly.

# Implementation
- [`Stage`](https://github.com/LabExT/LabExT/pull/11/files#diff-5105c6db4f4df249a2aa0186815847476c1e7c60e7537e151bdcc56c710f0b28) is implemented as AbstractClass. This means it cannot be initiated directly, only a child class can be initiated if it implements all abstract methods.
- [`Stage3DSmarAct`](https://github.com/LabExT/LabExT/pull/11/files#diff-2bfba1fac45d9432c531f8f22e9c6a7476e57e95e8c81a893b4fa28158ab5196) implements a SmarAct Stage using driver version 1. I suggest that the later goal should be to load the drivers (i.e. the PythonWrapper) here only, and use this class to perform all SmartAct specific operations.
- There are two different test files. One tests all base functionallity of `Stage3DSmarAct` and one tests the correct communication with the channels. Both test classes inherit from `SmarActTestCase`, which handles the mocking of the drivers. **Important**: All tests that also require mocking of the drivers must use this test case.

# Compatibility
This PR does not use the `Stage3DSmarAct ` class yet, `PiezoStage` is still [used](https://github.com/LabExT/LabExT/blob/main/LabExT/Movement/Mover.py#L154). Therefore LabExT should still work. After a test in the lab, I will create a PR to exchange the classes.

### DIFF
--- a/LabExT/Movement/Stage3DSmarAct.py
+++ b/LabExT/Movement/Stage3DSmarAct.py
@@ -10,7 +10,7 @@ import time
 import ctypes as ct
 from enum import Enum
 
-from LabExT.Movement.Stage import Stage
+from LabExT.Movement.Stage import Stage, StageError
 from LabExT.Utils import get_configuration_file_path
 
 sys_path_changed = False
@@ -253,7 +253,11 @@ class Stage3DSmarAct(Stage):
 
         # Movement
 
-        def move(self, diff: float, mode=MovementType.RELATIVE, wait_for_stopping=True) -> None:
+        def move(
+                self,
+                diff: float,
+                mode=MovementType.RELATIVE,
+                wait_for_stopping=True) -> None:
             """Moves the channel with the specified movement type by the value diff
 
             Parameters
@@ -290,7 +294,8 @@ class Stage3DSmarAct(Stage):
                     ct.c_int(int(self._to_nanometer(diff))),
                     0
                 ))
-            if wait_for_stopping: self._wait_for_stopping()
+            if wait_for_stopping:
+                self._wait_for_stopping()
 
         def _move_absolute(self, pos: float, wait_for_stopping: bool):
             self._stage._exit_if_error(
@@ -300,7 +305,8 @@ class Stage3DSmarAct(Stage):
                     ct.c_int(int(self._to_nanometer(pos))),
                     0
                 ))
-            if wait_for_stopping: self._wait_for_stopping()
+            if wait_for_stopping:
+                self._wait_for_stopping()
 
         def _humanize_status(self, status_code: int) -> str:
             if(status_code in self.STATUS_CODES):
@@ -598,7 +604,7 @@ class Stage3DSmarAct(Stage):
     def _raise_if_sensor_non_linear(self) -> None:
         for index, channel in self.channels.items():
             if not channel.is_sensor_linear:
-                raise RuntimeError(
+                raise StageError(
                     'Channel {} of stage {} has no supported linear sensor!'.format(
                         index.name, self.address))
         self.logger.debug("Linear x, y and z sensor present")
@@ -607,14 +613,13 @@ class Stage3DSmarAct(Stage):
         if(status == MCSC.SA_OK):
             return True
 
-        error_msg = ct.c_char_p()
-        MCSC.SA_GetStatusInfo(status, error_msg)
+        error_message = ct.c_char_p()
+        MCSC.SA_GetStatusInfo(status, error_message)
 
-        if(error_msg):
-            self.logger.error(
-                'MCS error: {}'.format(
-                    error_msg.value[:].decode('utf-8')))
+        if error_message:
+            error_message = 'MCSControl Error: {}'.format(
+                error_message.value[:].decode('utf-8'))
         else:
-            self.logger.error('MCS error: undefined error occurred.')
+            error_message = 'MCSControl Error: Undefined error occurred.'
 
-        return False
+        raise StageError(error_message)
--- a/LabExT/Movement/Stage3DSmarAct.py
+++ b/LabExT/Movement/Stage3DSmarAct.py
@@ -176,7 +176,7 @@ class Stage3DSmarAct(Stage):
                     self._stage.handle,
                     self._handle,
                     system_position)):
-                self._position = self._to_mircometer(system_position.value)
+                self._position = self._to_micrometer(system_position.value)
 
             return self._position
 
@@ -189,7 +189,7 @@ class Stage3DSmarAct(Stage):
                     self._stage.handle,
                     self._handle,
                     system_speed)):
-                self._speed = self._to_mircometer(system_speed.value)
+                self._speed = self._to_micrometer(system_speed.value)
 
             return self._speed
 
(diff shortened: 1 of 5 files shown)


marcoep:
Before I go into details for reviewing:
* Why doesn't the CI testing action get run for this PR?
* Please merge into the dev branch. The main branch should be only for released code.

maltewae:
- This was my mistake: In my PR to setup the CI I specified that the run gets only triggered on push, but I didn't push to the repo, since I am trying to merge my fork branch. I will change it, such that it will be triggered when someone pushed or opened a PR.
- Changed the base branch, I will rebase it in a second. :+1

maltewae:
I rebased the PR onto dev.

LabExT/Movement/Stage3DSmarAct.py
```diff
@@ -0,0 +1,610 @@
...
+
+    def _raise_if_sensor_non_linear(self) -> None:
+        for index, channel in self.channels.items():
+            if not channel.is_sensor_linear:
+                raise RuntimeError(
+                    'Channel {} of stage {} has no supported linear sensor!'.format(
+                        index.name, self.address))
+        self.logger.debug("Linear x, y and z sensor present")
+
+    def _exit_if_error(self, status: int) -> bool:
```
maltewae, line 596: Discussion point: I kept this method similar to [PiezaStage.py](https://github.com/LabExT/LabExT/blob/main/LabExT/Movement/PiezoStage.py#L151). Only a boolean return value I introduced, but we don't really "exit" the program here. Should we raise an exception here so that the program stops?

marcoep: Yes, I would suggest to immediately raise a RuntimeError with a telling error message.

LabExT/Movement/Stage3DSmarAct.py
```diff
@@ -0,0 +1,620 @@
...
+                ))
+            if wait_for_stopping: self._wait_for_stopping()
+
+        def _humanize_status(self, status_code: int) -> str:
+            if(status_code in self.STATUS_CODES):
+                return self.STATUS_CODES[status_code]
+
+            return "Unknown status code: " + str(status_code)
+
+        def _wait_for_stopping(self):
```
maltewae, line 311: I reintroduced this feature after testing it in the lab. It turns out that this is needed for the function `wiggle_z_axis_positioner` for example. If we don't wait until the stages have stopped, we change the speed in `wiggle_z_axis_positioner` while the stages are still moving.

LabExT/Movement/Stage3DSmarAct.py
maltewae: I agree. Will change it 👍🏼

LabExT/Movement/Stage3DSmarAct.py
```diff
@@ -0,0 +1,625 @@
...
+            while True:
+                time.sleep(0.05)
+                if self.status == MCSC.SA_STOPPED_STATUS:
+                    break
+
+        def _to_nanometer(self, um: int) -> int:
+            return um * 1e3
+
+        def _to_mircometer(self, nm: int) -> int:
```
marcoep, line 326: typo: its micrometer

LabExT/Movement/Stage3DSmarAct.py
```diff
@@ -0,0 +1,625 @@
...
+            if wait_for_stopping:
+                self._wait_for_stopping()
+
+        def _humanize_status(self, status_code: int) -> str:
+            if(status_code in self.STATUS_CODES):
+                return self.STATUS_CODES[status_code]
+
+            return "Unknown status code: " + str(status_code)
+
+        def _wait_for_stopping(self):
```
marcoep, line 317: I suggest adding an optional timeout feature here. If the movement takes longer than a specified amount of seconds, you can raise a RuntimeError here.

LabExT/Movement/Stage3DSmarAct.py
```diff
@@ -0,0 +1,625 @@
...
+            if mode.value not in MovementType._value2member_map_:
+                raise ValueError("Invalid movement mode {}".format(str(mode)))
+            self._movement_mode = mode
+
+        # Movement
+
+        def move(
+                self,
+                diff: float,
+                mode=MovementType.RELATIVE,
```
marcoep, line 259: I would make this a mandatory argument, so to force the programmer to specify it.

LabExT/Movement/Stage3DSmarAct.py
marcoep: also, the number in micrometers is of type float.

LabExT/Movement/Stage3DSmarAct.py
```diff
@@ -0,0 +1,625 @@
...
+
+            return "Unknown status code: " + str(status_code)
+
+        def _wait_for_stopping(self):
+            while True:
+                time.sleep(0.05)
+                if self.status == MCSC.SA_STOPPED_STATUS:
+                    break
+
+        def _to_nanometer(self, um: int) -> int:
```
marcoep, line 323: type of um argument is usually float

LabExT/Movement/Stage3DSmarAct.py
maltewae: I agree. But what would be a reasonable time to wait? Assuming the stage is moving at very low speed over a long distance, it would take longer for the stage to come to a stop. We could ask the user after a certain time through a dialog if he wants to stop the stage.

LabExT/Movement/Stage3DSmarAct.py
maltewae: Thanks! Will change it.

LabExT/Movement/Stage3DSmarAct.py
maltewae: Thanks! Will change it.

LabExT/Movement/Stage3DSmarAct.py
maltewae: Good point, I agree. Will change it 👍🏼

LabExT/Movement/Stage3DSmarAct.py
marcoep: So my idea was that `Stage.move_relative` and `Stage.move_absolute` have a timeout feature that you can enable (bool!). When enabled, Python would throw a RuntimeError if the stage takes longer for the required movement than we would expect based on the theoretical calculation with acceleration and speed + 20% overhead.

But since we are still just refactoring, dont worry about it, its fine.

marcoep (approved):

PR #15, opened 2021-11-24
[Movement] Find SmarAct systems

maltewae:
**Don't merge before https://github.com/LabExT/LabExT/pull/11**

# Goal
Add a classmethod to Stage3DSmarAct for finding all SmarAct stages.

# Approach
By invoking the driver method `SA_FindSystems`, we should be able to find all usb-connected stages.

# Compatibility
This will not break anything, since we aren't using the method yet.

- [x] I have tested the function in the Lab with hardware.

### DIFF
--- a/LabExT/Movement/Stage.py
+++ b/LabExT/Movement/Stage.py
@@ -14,11 +14,16 @@ class StageError(RuntimeError):
 
 
 class Stage(ABC):
+    _logger = logging.getLogger()
+
+    @classmethod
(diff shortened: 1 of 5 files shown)


marcoep:
Review done, but please change target branch to LabExT:dev.

maltewae:
I rebased onto dev. Just need your approval to merge.

marcoep (approved):

PR #17, opened 2021-11-26
[Movement] Apply movement refactoring

maltewae:
# Goal
Apply movement refactoring introduced in #11 

# Approach
Deleted PiezoStage class and use Stage3DSmarAct instead. Use Stage3DSmarAct to find all stages in Mover class.

# Compatibility
- [x] Tested in Lab with hardware


marcoep (commented):
Good stuff!

A approve as soon as this is tested in lab.

marcoep (approved):

PR #18, opened 2021-11-27
Feature: Limit y-axis scaling in live viewer

marcoep:
# Goal
Add live-viewer feature: limit y-axis scaling to minimum.

When using the live viewer, the y-axis currently is scaled to always show the total y-axis range + 20%. This could mean that the y-axis spans about 0.2 dB during optical coupling, which is too small to make sense. Ideally, we want to lower-limit to 1 to 2 dB y axis range.

# Approach
Adding the property `min_y_axis_span` to PlotControl plus the scale limiting in `__handle_scaling__`. Further, the LiveViewer controller sets the attribute on clicking the apply settings button.

# Compatibility
No compatibility issues expected, its just a small added feature to PlotControl which is disabled by default.

### DIFF
--- a/LabExT/View/Controls/PlotControl.py
+++ b/LabExT/View/Controls/PlotControl.py
@@ -180,6 +180,25 @@ class PlotControl(Frame):
         self._axis_limit_margin_fraction = h
         self.__handle_scaling__()
 
+    @property
+    def min_y_axis_span(self):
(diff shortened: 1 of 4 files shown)


marcoep:
@maltewae Maybe you also want to have a look at this code and give me pointers on how to improve.

boris-vukovic (approved):

maltewae:
Looks good to me! 🚀

PR #19, opened 2021-12-01
[Movement] Decorator for safe access to the stage methods

maltewae:
# Goal
Raise Runtime error when trying to invoke a stage method which requires an active stage connection and loaded drivers.

# Approach
Introducing a decorator for functions which needs stage driver and stage connection. A runtime error is triggered if one of these conditions is not met.
I also added tests for the decorator.

This a follow-up PR of #17


maltewae:
Tested it in labor. It will raises a Runtime error on stage functions, if stage is not connected. Otherwise, it works like before.

PR #21, opened 2021-12-11
[Movement] (1/3) MovementWizard: Manage connection to stages

maltewae:
# Goal
Add a new MovementWizard and Mover implementation.
This PullRequest focuses on establish connections to stages.

# Approach and Implementation
- MoverClass: Added a new class named MoverNew. So that for the time being we can use both movers at the same time. This class manages a list of stages, which are currently plugged in but not necessarily connected. These stages will be found by a find_stages method. Each stage implementation in LabExT (e.g Stage3DSmarAct) should provide this function and should return all available stages of its kind.
The Mover provides function to disconnect and connect to all or to selected stages.

- MoverTest: I added tests to test the Mover functions. The tests assume that we have a correct Stage implementation, so the test mock the out and performs based on the cases: No available stages, only connected stages, only disconnected stages and partial connected stages.

- MovementWizardController: Handles everything to connect and disconnect stages and to find all available stages
- MovementWizardView: Handles representation of model (MoverClass) and interacts with controller MovementWizardController

# Compatibility
This PR does introduce new code which is not actively used.

### DIFF
--- a/LabExT/ExperimentManager.py
+++ b/LabExT/ExperimentManager.py
@@ -18,6 +18,7 @@ from LabExT.Experiments.StandardExperiment import StandardExperiment
 from LabExT.Instruments.InstrumentAPI import InstrumentAPI
 from LabExT.Instruments.ReusingResourceManager import ReusingResourceManager
 from LabExT.Movement.Mover import Mover
+from LabExT.Movement.MoverNew import MoverNew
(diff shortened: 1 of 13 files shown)


marcoep:
@maltewae I suggest to merge this to a separate movement-refactoring branch before we clutter the dev branch with half-done refactorings.

maltewae:
@marcoep Added a branch `dev_movement` and changed the base branch to it.

PR #25, opened 2022-01-11
DummyMeas.py: now sleeps in any case

marcoep:
See title: small bugfix in dummy meas


marcoep:
@boris-vukovic pls review :)

boris-vukovic (approved):

PR #26, opened 2022-01-11
bugfix: ignore Keysight mainframe dB calculation error 

marcoep:
Apparently, on older mainframe (816xB) firmwares a -261:StatUnitCalculationError will appear on very low optical input powers. For the explanation, see this excerpt of the 8164B Programming Guide: 

image

In this PR, we simply ignore the error, as its annoying and interrupts live viewing and measurements. We can savely ignore it, as too low optical powers get recorded as NaN anyhow. (Except on the N77xx models, which report a -200dBm value for too low powers.)

### DIFF
--- a/LabExT/Instruments/LaserMainframeKeysight.py
+++ b/LabExT/Instruments/LaserMainframeKeysight.py
@@ -43,7 +43,7 @@ class LaserMainframeKeysight(Instrument):
 
     """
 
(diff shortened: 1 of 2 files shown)


marcoep:
@boris-vukovic pls review :)

boris-vukovic (approved):

PR #31, opened 2022-01-19
Add py3.7 and py3.9 to test pipeline

maltewae:


marcoep:
@maltewae pls also add 3.7

Will do another dependency upgrade for python version 3.7 compatible and available packages.

marcoep:
@maltewae I fixed the mock tests to be python 3.7 compatible but, frustratingly, I cannot seem to find the last bug caused in `test_disconnect_unsuccessfully`. All tests run flawlessly on my Windows 10 machine using Python 3.7, so it could be a ubuntu thing, as we use this OS for the tests? Could I ask you to double check with your mac?

maltewae:
@marcoep If would suggest to change this line https://github.com/LabExT/LabExT/blob/d3aadcdaa7b8bc0126a3310516f8da659f70b3b0/LabExT/Tests/Movement/Stage3DSmarAct/Base_test.py#L108 to `self.mcsc_mock.SA_CloseSystem.assert_called_with`.
I think the problem is that after raising the `StageError` the `Stage3DSmarAct `-object will be deleted and the `del` method will be invoked. And this method will again call `disconnect()` https://github.com/LabExT/LabExT/blob/d3aadcdaa7b8bc0126a3310516f8da659f70b3b0/LabExT/Movement/Stage.py#L52. Thus we have more than one invocations.

It's just strange that the other tests go through. They should also fail. I'll take a look, but for now I would take out the condition that it should only be called once.

marcoep:
@maltewae You are right, please do so. In the end, it does not matter if the stage has been closed once or three times, as long as its more than 0 times :)

maltewae:
Interesting, the tests also all run through on the Mac without any problems. 🤔

maltewae:
If you want, we can also change the test environment to `windows-latest`, as our lab equipment all runs on Windows.

marcoep:
@maltewae And what is this error now?

PR #36, opened 2022-02-03
[Movement] and [Tests]: Stage module and new Tests

maltewae:
# TL;DR: This PullRequest reorganises stages and tests in LabExT:
- Entrypoint for tests in LabExT
- New methods for Stage to find Stages.
- Method to load SmarAct drivers
- More stability for SmarAct tests (fixes https://github.com/LabExT/LabExT/pull/31)
- DummyStage added

# All changes in detail:

## 1. Entry point for LabExT tests:
The file [runtests](https://github.com/maltewae/LabExT/blob/5b979ae5c6060ab54b1f52bd585752a63394bd87/LabExT/Tests/runtests.py) is the entry point for all tests. In the future, the test suite must be called with `python -m LabExT.Tests.runtests` instead of `pytest`. In the background, `runtests` calls `pytest.main()` to run the tests. `runtests` takes the optional `--laboratory_tests` which tells the test-suite to also run tests that require laboratory equipment. In CI these are **skipped**.

### How do I mark tests that require laboratory equipment?
In `LabExT/Tests/Utils.py` there is a decorator `mark_as_laboratory_test` which can be used to mark a `TestCase` as a laboratory test:
```python
@mark_as_laboratory_test
class Test_PiezoStage(unittest.TestCase):
    ...
```

**The tests `ExampleInstrument_test`, `OpticalSpectrumAnalyzerAQ6370C_test`, `PowerMeterGenericKeysight_test`, `DummyMeas_test`, `InsertionLossSweep_test`, `ReadOSA_test` and `PiezoStage_test` were adjusted accordingly and were included in the test suite.**

## 2. `DummyStage`
Creation of a [DummyStage](https://github.com/maltewae/LabExT/blob/5b979ae5c6060ab54b1f52bd585752a63394bd87/LabExT/Movement/Stages/DummyStage.py). This is helpful for testing LabExT.

## 3. Restruture of `Stage3DSmarAct`-Test
The main goal was to make the tests more stable and compatible with Python 3.7 (see https://github.com/LabExT/LabExT/pull/31).

### The following changes were made:
1. deletion of `SmarActTestCase`, `ChannelTest` and `BaseTest`. These tests are now all in one file `Stage3DSmarAct_test`. Why: The `ChannelTest` and `BaseTest` files were a bit misleading, as it was not directly understood what was specifically tested.
2. the mocking of the stage driver has been changed. For this I created an interface for the PythonWrapper of the driver ([MCSControlInterface.py](https://github.com/maltewae/LabExT/blob/5b979ae5c6060ab54b1f52bd585752a63394bd87/LabExT/Tests/Fixtures/MCSControlInterface.py)). This does **not** contain any implementation logic but only the signatures of the functions. Why: With the help of the autospec function of unittest-mocks a mock object is created that only allows function calls with given signature which are defined the module `MCSControlInterface`. This ensures that during testing the call of e.g `mcsc_mock.SA_FindSystems(...)` succeeds (since defined in the interface) and `mcsc_mock.some_typo(...)` fails (since not defined). This increases the test security significantly.

**Otherwise, the tests have remained the same.** 

# 4. Minor changes have been made:
1. renaming of `find_stages` to `find_stage_addresses`
2. Added a method `find_stage_classes` to `Stage` to find all stage classes which inherits from `Stage`
3. A method `find_available_stages` has been added to `Stage` to return a list of `Stage` objects, where each object represents a stage connected to the computer. 
4. adding a `load_driver` function for SmarAct stages. This calls the `DriverPathDialog` to specify a driver path.


LabExT/Tests/Utils.py
```diff
@@ -26,6 +27,13 @@ def pump_events(self):
             pass
 
 
+def mark_as_laboratory_test(klass):
```
marcoep, line 30:
```suggestion
def mark_as_laboratory_test(class):
```

LabExT/Tests/Utils.py
```diff
@@ -26,6 +27,13 @@ def pump_events(self):
...
+    """
+    Decorator to mark test as laboratory tests. These will be exclude, when run on CI.
+    """
+    return pytest.mark.skipif(not pytest.run_laboratory_tests, reason="skip tests that require laboratory equipment.")(klass)
```
marcoep, line 34:
```suggestion
    return pytest.mark.skipif(not pytest.run_laboratory_tests, reason="skip tests that require laboratory equipment.")(class)
```

LabExT/Tests/Utils.py
```diff
@@ -26,6 +27,13 @@ def pump_events(self):
             pass
 
 
+def mark_as_laboratory_test(klass):
+    """
+    Decorator to mark test as laboratory tests. These will be exclude, when run on CI.
```
marcoep, line 32:
```suggestion
    Decorator to mark test as laboratory tests. These will be excluded, when run on CI.
```

LabExT/Tests/Measurements/InsertionLossSweep_test.py
```diff
@@ -32,7 +33,8 @@ def check_ILsweep_data_output(test_inst, data_dict, params_dict):
                                params_dict['wavelength stop'].value))
 
 
-class Test_InsertionLossSweep(unittest.TestCase):
+@mark_as_laboratory_test
```
marcoep, line 36: This test can be run with the supplied simulator classes and does NOT need any instruments attached. No need to classify it as a lab test.

```suggestion
```

LabExT/Tests/Fixtures/MCSControlInterface.py
```diff
@@ -0,0 +1,601 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+LabExT  Copyright (C) 2022  ETH Zurich and Polariton Technologies AG
+This program is free software and comes with ABSOLUTELY NO WARRANTY; for details see LICENSE file.
```
marcoep, line 5: Not sure if this is correct. Most of these constants are supplied by SmarAct which have a proprietary license on their drivers.

We can argue okay, we are not supplying the drivers but merely interface with them. But I would add an acknowledgment to SmarAct here.

LabExT/Movement/Stages/Stage3DSmarAct.py
```diff
@@ -53,6 +56,42 @@ class Stage3DSmarAct(Stage):
...
     """
 
+    driver_loaded = MCS_LOADED
+
+    @classmethod
+    def load_driver(cls, parent) -> bool:
+        """
+        Loads driver for SmarAct by open a dialog to specifiy the driver path. This method will be invoked by the MovementWizardController.
+        """
+        driver_path_dialog = DriverPathDialog(
```
marcoep, line 66: Maybe check here if driver is already loaded before opening GUI element.

marcoep (changes requested):
Good stuff. But I fail to see how this should fix #31? Could you explain, please?

LabExT/Tests/Utils.py
maltewae: Changed it to `cls`, since `class` is a reserved keyword.

LabExT/Tests/Utils.py
maltewae: Changed it to cls, since class is a reserved keyword.

LabExT/Tests/Utils.py
maltewae: Fixed 👍🏼

LabExT/Movement/Stages/Stage3DSmarAct.py
maltewae: Good idea 👍🏼  I changed it.

LabExT/Tests/Measurements/InsertionLossSweep_test.py
maltewae: ups, I overlooked that. Now they are included on CI

LabExT/Tests/Fixtures/MCSControlInterface.py
maltewae: I agree, I wasn't sure if we were allowed to do that. If it's because of the constants, we can also remove them from the interface. Otherwise it is useful to know during testing which methods the `MCSControl_PythonWrapper` module provides. 😄

maltewae:
Why it works now with Python 3.7 (I think 😄 ): Before I mocked the driver module globally, now I create a mock for each unit test. I think before the counter on the method calls (used for `asset_called_once`, `assert_called` etc.) was affected by this between the tests. In the `teardown` method of each test I now set `stage.connected` to `False`, so that the `__del__ `hook is not triggered.

marcoep (approved):
noice!

LabExT/Tests/Fixtures/MCSControlInterface.py
marcoep: eh... lets keep it for now, Ill ask SmarAct via Email if this is okay.

marcoep:
good to merge

marcoep:
good to merge! @boris-vukovic pls :)

PR #44, opened 2022-02-16
[Movement] (2/3): Axes Calibration for safe relative Movement

maltewae:


LabExT/Movement/Calibration.py
```diff
@@ -119,3 +206,45 @@ def connect_to_stage(self) -> bool:
...
+    def wiggle_axis(
+            self,
+            axis: Axis,
+            axes_rotation: Type[AxesRotation],
+            wiggle_distance=1e3,
+            wiggle_speed=1e3):
+        """
+        Wiggles the requested axis positioner in order to enable the user to test the correct direction and axis mapping.
+        """
+        x_stage, y_stage, z_stage = axes_rotation.chip_to_stage(np.array([
```
marcoep, line 233: Wouldn't it make sense to access `self._axes_rotation` here?

Also: wouldn't it make sense to check if the current `_state` is COORDINATE_SYSTE_FIXED? The method is currently written such that it only works in this state.

LabExT/Movement/Calibration.py
```diff
@@ -45,6 +74,62 @@ def __str__(self) -> str:
...
+        A matrix is a permutation matrix if the sum of each row and column is exactly 1.
+        """
+        abs_matrix = np.absolute(self._matrix)
+        return (
+            abs_matrix.sum(
+                axis=0) == 1).all() and (
+            abs_matrix.sum(
+                axis=1) == 1).all()
+
+    def chip_to_stage(self, chip_coordinate):
```
marcoep, line 120: Is my assumption correct that this function is only correct for relative coordinates? If yes please update docstring accordingly.

You don't need to change anything, but we should keep in mind that for the full calibration it can come in handy to have `chip_to_stage` (and `stage_to_chip`) functions available for both, relative and absolute coordinates.

LabExT/View/StageCalibration/StageCalibrationView.py
```diff
@@ -0,0 +1,197 @@
...
+        for calibration in self.mover.calibrations.values():
+            self._current_axes_rotations[calibration] = AxesRotation()
+
+            axes_vars = {}
+            for chip_axis in Axis:
+                axes_vars[chip_axis] = StringVar(
+                    self.parent, self.STAGE_AXIS_OPTIONS[(Direction.POSITIVE, chip_axis)])
+                axes_vars[chip_axis].trace(
+                    W,
+                    lambda *_,
```
marcoep, line 60: What is this kind of fancy syntax? 😄

marcoep (approved):
Good stuff. Just some general remarks but ready to merge :)

LabExT/Movement/Calibration.py
maltewae: Mhm good point. I wrote the `wiggle_axis` method this way to test a given `axes_rotation` (that's why as an argument to the method) before it is saved (i.e. `self._axes_rotation` is set.) That's why I deliberately didn't use `self._axes_rotation` and don't check `COORDINATE_SYSTEM_FIXED`.
`COORDINATE_SYSTEM_FIXED` is only set when the method `fix_coordinate_system` is called and a rotation is set for `self._axes_rotation`.

Is this too confusing? My idea was that `self._axes_rotation` is not set and used for general relative movements until you are satisfied with the rotation.

LabExT/View/StageCalibration/StageCalibrationView.py
maltewae: The trace method for variables calls the method with arguments that I don't need. `*_` is like `*args`, except that the `_` implies that the argument is not used.

LabExT/Movement/Calibration.py
marcoep: Ahh, okay. This makes perfectly sense and I agree with that design pattern.

PR #45, opened 2022-02-18
[Movement] (3/5): Single Point für safe absolute (approx) movement

maltewae:
# Movement Update: Single Point Fixation
This PR adds another step to the calibration of the stages. For each stage a coordinate pair of chip and stage coordinate is defined, which can be used to calculate a translation. Together with the axis-calibration it is possible to move approximate absolute in chip-coordinates.

## All changes in detail:
- `Transformation`: Is an abstract base class, so this cannot be initialised directly and provides an interface for all transformations in LabExT. This requires that the functions `chip_to_stage`, `stage_to_chip` and `is_valid` are defined.
- `CoordinatePairing`: `NamedTuple`. Quadruple of calibration (aka stage), stage coordinate (list), device and chip coordinate.
- `SinglePointFixation` is a transformation. It stores a chip and stage cooridnate and calculates an offset from these. This is added to the coordinate at `chip_to_stage` and `stage_to_chip`.
- `CoordinatePairingsWindow` is a simple window that can be passed one or two calibrations, where one must be an input and the other an output stage. After selecting a device and asking to move the stages to the ports, 1 or 2 pairings are created (depending on how many calibrations have been passed).
- `StageCalibrationView` and `StageCalibrationController` have been adapted for the new step
- Tests for all new methods
Screenshot 2022-02-21 at 10 57 15
Screenshot 2022-02-21 at 10 57 31
Screenshot 2022-02-21 at 10 57 39
Screenshot 2022-02-21 at 10 57 52

### DIFF
--- a/LabExT/Movement/Transformations.py
+++ b/LabExT/Movement/Transformations.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+LabExT  Copyright (C) 2022  ETH Zurich and Polariton Technologies AG
+This program is free software and comes with ABSOLUTELY NO WARRANTY; for details see LICENSE file.
+"""
+
+from abc import ABC, abstractmethod
+from typing import NamedTuple, Type
+import numpy as np
+
+
+class Transformation(ABC):
+    """
+    Abstract interface for transformations.
+    """
+
+    @abstractmethod
+    def __init__(self) -> None:
+        pass
+
+    @property
+    @abstractmethod
+    def is_valid(self):
+        """
+        Returns True if the transformation is valid.
+        """
+        pass
+
+    @abstractmethod
+    def chip_to_stage(self, chip_coordinate):
+        """
+        Transforms a coordinate in chip space to stage space.
+        """
+        pass
+
+    @abstractmethod
+    def stage_to_chip(self, stage_coordinate):
+        """
+        Transforms a coordinate in stage space to chip space.
+        """
+        pass
+
+
+class CoordinatePairing(NamedTuple):
+    calibration: object
+    stage_coordinate: list
+    device: object
+    chip_coordinate: list
+
+
+class SinglePointFixation(Transformation):
+    """
+    Performs a translation based on a fixed single point.
+    """
+
+    def __init__(self) -> None:
+        self._chip_coordinate = None
+        self._stage_coordinate = None
+        self._offset = None
+
+    def __str__(self) -> str:
+        if self._offset is None:
+            return "No single point fixed"
+
+        return "Stage-Coordinate {} fixed with Chip-Coordinate {}".format(
+            self._stage_coordinate, self._chip_coordinate)
+
+    @property
+    def is_valid(self):
+        """
+        Returns True if single point transformation is defined.
+        """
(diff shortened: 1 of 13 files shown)


LabExT/View/Movement/CoordinatePairingsWindow.py
```diff
@@ -0,0 +1,271 @@
...
+
+from tkinter import Frame, Toplevel, Button, Label, messagebox, LEFT, RIGHT, TOP, X, BOTH, DISABLED, FLAT, NORMAL, Y
+from LabExT.View.Controls.CustomFrame import CustomFrame
+from LabExT.View.Controls.DeviceTable import DeviceTable
+
+from LabExT.Movement.Stage import StageError
+import LabExT.Movement.Transformations as Transformations
+
+
+class CoordinatePairingsWindow(Toplevel):
```
marcoep, line 16: I am wondering where you enforce that only one coordinate pairing (as its single point fixed) can be done. Is this inherent to this window, so it will only allow a single point per attached stage?

LabExT/View/Movement/CoordinatePairingsWindow.py
```diff
@@ -0,0 +1,271 @@
...
+                relief='sunken').pack(
+                side=LEFT,
+                padx=5)
+        else:
+            Label(
+                pairing_frame,
+                text="No device selected",
+                foreground="#FF3333"
+            ).pack(side=LEFT, padx=5)
+
```
marcoep, line 256: A cool feature here would be to have a display of the current stage coordinates.

E.g. a "will be paired with stage coordinates: XXX YYY ZZZ" label or so.

This suggestion needs a background-running routine which reads and updates the coordinates from the stage accordingly.

Lets talk about it in the next meeting, don't spend time implementing it just yet :)

marcoep (approved):

LabExT/Movement/Transformations.py
```diff
@@ -0,0 +1,104 @@
...
+        return np.array(chip_coordinate) + self._offset
+
+    def stage_to_chip(self, stage_coordinate):
+        """
+        Translates stage coordinate to chip coordinate
+        """
+        if not self.is_valid:
+            raise RuntimeError("Cannot translate with invalid fixation. ")
+
+        return np.array(stage_coordinate) + self._offset
```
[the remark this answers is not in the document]
maltewae: uhh embarrassing, good catch! 😳  I fixed it and written two tests that checks the following:
```python
input = chip_to_stage(stage_to_chip(input))
input = stage_to_chip(chip_to_stage(input))
```

LabExT/View/Movement/CoordinatePairingsWindow.py
maltewae: In fact, I have such a feature in a private branch, so it's easy for me to add it. 😄 Will be up in a minute.

maltewae: Screenshot 2022-02-22 at 17 15 18

maltewae:
Any updates to this? 😄

marcoep (approved):

maltewae:
Rebased to `movement/main` after merge of #47

PR #47, opened 2022-02-22
[Movement][Fix]: Wait for Stage Stop

maltewae:
As discussed in yesterday's meeting, we are waiting in each channel of the Stage for it to enter `STOPPED` status. This causes the Stage to move each axis in turn, rather than in all directions at once.

This PR moves the `wait_for_stopping` feature out of the channel and into the Stage, which manages all channels simultaneously. Tests have been added to test the general behavior.


LabExT/Movement/Stages/Stage3DSmarAct.py
```diff
@@ -655,7 +638,7 @@ def get_current_position(self) -> list:
 
     @assert_driver_loaded
     @assert_stage_connected
-    def move_relative(self, x, y):
+    def move_relative(self, x, y, wait_for_stopping: bool = True):
```
delmedico, line 641: I think the lift_stage() and lower_stage() need the the wait_for_stopping flag as well put to True as default

maltewae: Yep, you're right. I have adjusted the method `lower_stage` and `lift_stage`. These now call the `move_relative` method (with `wait_for_stopping` flag). I added tests for this.

delmedico (approved):
very nice!

maltewae:
@marcoep Is this good to merge?

LabExT/Movement/Stages/Stage3DSmarAct.py
```diff
@@ -738,3 +737,12 @@ def _raise_if_sensor_non_linear(self) -> None:
                     'Channel {} of stage {} has no supported linear sensor!'.format(
                         index.name, self.address))
         self._logger.debug("Linear x, y and z sensor present")
+
+    def _wait_for_stopping(self, deplay=0.05):
```
marcoep, line 741:
```suggestion
    def _wait_for_stopping(self, delay=0.05):
```

LabExT/Movement/Stages/Stage3DSmarAct.py
```diff
@@ -738,3 +737,12 @@ def _raise_if_sensor_non_linear(self) -> None:
...
+        """
+        Blocks until all channels have 'SA_STOPPED_STATUS' status.
+        """
+        while True:
+            time.sleep(deplay)
```
marcoep, line 746:
```suggestion
            time.sleep(delay)
```

marcoep (approved):
Just a typo :)

LabExT/Movement/Stages/Stage3DSmarAct.py
maltewae: Fixed 👍🏼

LabExT/Movement/Stages/Stage3DSmarAct.py
maltewae: Fixed 👍🏼

PR #48, opened 2022-03-21
Adding a full click-through unittest of running an InsertionLoss measurement with simulated instruments

marcoep:
This PR tries to add a full click-through unittest of the basic InsertionLossSweep measurement within the tkinter GUI.

### DIFF
--- a/.github/workflows/labext-ci.yml
+++ b/.github/workflows/labext-ci.yml
@@ -18,6 +18,6 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install tox tox-gh-actions
+        pip install -r requirements_testing.txt
     - name: Running tests
       run: tox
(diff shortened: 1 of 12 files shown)


marcoep:
@boris-vukovic ready for review!

boris-vukovic (changes requested):
requirements.txt is missing package pytest to run the test

marcoep:
@maltewae What do you think about Boris' proposition? How is it usually solved in other packages that you include the packages for testing in requriements.txt?

Should we make an extra requirements_testing.txt? Or should we just include pytest in the "live" requirements.txt?

maltewae:
@marcoep I would suggest to create a second dependency file for the testing environment to keep the production deps small. We could also use the [tox.ini](https://github.com/LabExT/LabExT/blob/main/tox.ini) for that, but then we would always have to run the tests locally via Tox.

marcoep:
@boris-vukovic should be ready :) we moved the testing requirements to its separate file

boris-vukovic (approved):

PR #59, opened 2022-05-12
Movement Calibration [4/7] Position of Stage

maltewae:
Adds a method to read out the position of the stage in stage coordinates (no change) or in chip coordinates. To read it in chip coordinates, you need at least a single point offset transformation or a Kabsch rotation, the later being preferred.

### DIFF
--- a/LabExT/Movement/Calibration.py
+++ b/LabExT/Movement/Calibration.py
@@ -38,7 +38,7 @@ def assert_minimum_state_for_coordinate_system(
     """
     def assert_state(func):
         @wraps(func)
-        def wrap(calibration: Type[Calibration], *args, **kwargs):
+        def wrap(calibration: Type["Calibration"], *args, **kwargs):
             if calibration.coordinate_system is None:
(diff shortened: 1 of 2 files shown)


LabExT/Movement/Calibration.py
```diff
@@ -311,3 +311,45 @@ def determine_state(self, skip_connection=False) -> None:
...
 
         self._state = State.FULLY_CALIBRATED
+
+    #
+    #   Position method
+    #
+
+    @assert_minimum_state_for_coordinate_system(
+        stage_coordinate_system=State.CONNECTED,
+        chip_coordinate_system=State.SINGLE_POINT_FIXED)
```
marcoep, line 321: Do I get it right that if this function is called within "stage_coordinate" context manager, then you don't care what the chip_coordinate_system state is, right?

maltewae: Yes, the state is selected depending on the current coordinate system.

PR #74, opened 2022-05-23
[Movement] Path Planning: Stage Polygons

maltewae:
Adds a base class to define stage polygons in LabExT for path planning.

The first polygon is for a DC Probe, which is a rectangular polygon defined by fiber length, fiber radius and safety distance.

### DIFF
--- a/LabExT/Movement/Calibration.py
+++ b/LabExT/Movement/Calibration.py
@@ -15,7 +15,7 @@ import numpy as np
 from LabExT.Movement.config import DevicePort, Orientation, State, Axis, Direction
 from LabExT.Movement.Transformations import StageCoordinate, ChipCoordinate, CoordinatePairing, SinglePointOffset, AxesRotation, KabschRotation
 from LabExT.Movement.Stage import StageError
-
+from LabExT.Movement.PathPlanning import StagePolygon, DCProbe
 
 if TYPE_CHECKING:
     from LabExT.Movement.Stage import Stage
@@ -70,6 +70,8 @@ class Calibration:
     def __init__(self, mover, stage, orientation, device_port) -> None:
         self.mover = mover
         self.stage: Type[Stage] = stage
+
+        self.stage_polygon: Type[StagePolygon] = DCProbe(orientation)
 
         self._state = State.CONNECTED if stage.connected else State.UNINITIALIZED
         self._orientation = orientation
--- a/LabExT/Movement/Calibration.py
+++ b/LabExT/Movement/Calibration.py
@@ -15,7 +15,7 @@ import numpy as np
 from LabExT.Movement.config import DevicePort, Orientation, State, Axis, Direction
 from LabExT.Movement.Transformations import StageCoordinate, ChipCoordinate, CoordinatePairing, SinglePointOffset, AxesRotation, KabschRotation
 from LabExT.Movement.Stage import StageError
-from LabExT.Movement.PathPlanning import StagePolygon, DCProbe
+from LabExT.Movement.PathPlanning import StagePolygon, SingleModeFiber
 
 if TYPE_CHECKING:
     from LabExT.Movement.Stage import Stage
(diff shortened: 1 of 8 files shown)


LabExT/Movement/Calibration.py
```diff
@@ -71,6 +71,8 @@ def __init__(self, mover, stage, orientation, device_port) -> None:
         self.mover = mover
         self.stage: Type[Stage] = stage
 
+        self.stage_polygon: Type[StagePolygon] = DCProbe(orientation)
```
maltewae, line 74: For now it is hard coded, different polygons can be selected later in the `StageSetup` wizard.

LabExT/Movement/PathPlanning.py
```diff
@@ -0,0 +1,128 @@
...
+
+        Parameters
+        ----------
+        position : ChipCoordinate
+            Current position of the stage in Chip-Coordinates
+
+        Raises
+        ------
+        ValueError
+            If no outline is defined for the given orientation.
```
marcoep, line 110: irrelevant now

LabExT/Movement/PathPlanning.py
```diff
@@ -0,0 +1,128 @@
...
+        y_max = position.y + safe_fiber_radius
+
+        if self.orientation == Orientation.LEFT:
+            x_min -= self.FIBER_LENGTH
+        elif self.orientation == Orientation.RIGHT:
+            x_max += self.FIBER_LENGTH
+        elif self.orientation == Orientation.BOTTOM:
+            y_min -= self.FIBER_LENGTH
+        elif self.orientation == Orientation.TOP:
+            y_max += self.FIBER_LENGTH
```
marcoep, line 126: should you add an `else` case here and raise a ValueError in case orientation is something else?

marcoep (approved):

LabExT/Movement/PathPlanning.py
```diff
@@ -0,0 +1,128 @@
...
+            Current position of the stage in Chip-Coordinates
+        mesh_x : np.ndarray
+            X values of the meshgrid
+        mesh_y : np.ndarray
+            Y values of the meshgrid
+        """
+        x_min, x_max, y_min, y_max = self._create_outline(position)
+        return np.logical_and(
+            np.logical_and(x_min <= mesh_x, mesh_x <= x_max),
+            np.logical_and(y_min <= mesh_y, mesh_y <= y_max))
```
marcoep, line 91: If the used gridsize in the meshgrid is too large, it might be that the obstacle completely falls between mesh points, in which case the obstacle is not seen from the potential field algorithm. Please make sure that an obstacle always has at minimum one point assigned.

Also, you could artificially "enlargen" the obstacle by `gridsize/2 + epsilon` s.t. it is guaranteed that we hit meshgrid points?

LabExT/Movement/PathPlanning.py
maltewae: Now it is raising a valueError if no outline is defined for a given orientation.

LabExT/Movement/PathPlanning.py
maltewae: Added

LabExT/Movement/PathPlanning.py
maltewae: I have changed the following: The method `stage_in_meshgrid` is now also passed the grid size. With this, each obstacle can calculate a polygon that is safely in the grid. For the `SingleModeFiber` I did the following: If, for example, the absolute distance between `x_max` and `x_min` is less than or equal to the grid size, I add half the grid size plus 10 as an epsilon to `x_max`, and subtract this value for `x_min`.

Further, I want to introduce to the path algorithm the following: If the stage -position is in the grid, `stage_in_meshgrid` must return a mask with at least one true-value. When the obstacles are created we can check if this is the case. If not, we throw a RunTime error.

marcoep: Very good suggestion for an addition with the minimum-one-enabled obstacle :)

PR #78, opened 2022-05-25
[Movement] Move to device

maltewae:
Move to device with path planning.


LabExT/Movement/MoverNew.py
```diff
@@ -442,6 +445,39 @@ def _lift_lower_stages(lift):
         finally:
             _lift_lower_stages(-self.z_lift)
 
+    @assert_connected_stages
+    def move_to_device(self, chip: Type[Chip], device_id: Any):
```
marcoep, line 449: why not make the prototype directly: `move_to_device(self, device: Type[Device])`?

LabExT/Movement/MoverNew.py
```diff
@@ -442,6 +445,39 @@ def _lift_lower_stages(lift):
...
+        """
+        device = chip._devices.get(device_id)
+        if device is None:
+            raise MoverError(
+                f"No device was found for the given chip for the given device ID {device_id}.")
+
+        movement_commands = {}
+        if self.input_calibration:
+            movement_commands[self.input_calibration.orientation] = device.input_coordinate + \
+                ChipCoordinate(z=self.z_lift)
```
marcoep, line 473: Why do you add the z_lift coordinate to the target coordinate to move to? This does not make any sense as we want to exactly arrive at coordinates (x, y, 0), no?

z-lift is only temporary during movement

marcoep, line 473: Why do you add the z_lift coordinate to the target coordinate to move to? This does not make any sense as we want to exactly arrive at coordinates (x, y, 0), no?

z-lift is only temporary during movement

marcoep (changes requested):

LabExT/Movement/MoverNew.py
maltewae: Good point, I find the interface suboptimal as well.
The device class unfortunately has no reference to the chip (so can't do `device.chip` or something else) and I didn't want to change the device class as it is being customized by Nino at the moment. Would suggest that we adjust that when the new device implementation is in.

LabExT/Movement/MoverNew.py
maltewae: I added this to make the Stage go to the Device point (x,y) in the plane of the Z-Lift. In the current implementation of the avoidance algorithm the z-cooridnate is kept constant anyway, but I wanted to make sure that the stage moves like in a rectangle. First z-lift up, then on that plane to the new x-y value, then z-lift down.

PR #91, opened 2022-07-20
Get rid of Python 3.7 backwards compatibility

marcoep:
# Goal
Speed up automated testing by ditching the Python 3.7 backwards compatibility. We support Python 3.8 and 3.9, which should be fine.

# Implementation
Note that the dependencies in requirements.txt are only Python 3.8 compatible starting from here. So when trying to install LabExT in a 3.7 environment, the required packages will likely not be found.

# Compatibility
Backwards compatibility to Python 3.7 is dropped. Everyone running LabExT in Python 3.7 will need to update their Python installation.


marcoep:
@delmedico @boris-vukovic 

So, during our last meeting, @maltewae and I were thinking of ditching Python 3.7 backwards compatibility. It saves a bit of time during testing and we can potentially make use of some newer language features.

Please let me know in any case how you feel about that and especially if you think its too early, e.g. regarding Polariton internal usage?

maltewae (approved):

delmedico (approved):
Thank you @marcoep. I agree with these changes. We do not rely on any Python 3.7. I think it's a good idea to get rid of it in the testing.

PR #93, opened 2022-07-26
[Movement]: Move relative method

maltewae:
Implements move relative in new Mover.
Relative movements are performed without collision avoidance. When multiple stages are requested to move, they move clockwise unless otherwise requested.

Tests for this behavior have been added.

### DIFF
--- a/LabExT/Movement/MoverNew.py
+++ b/LabExT/Movement/MoverNew.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, Tuple, Type, List
 from functools import wraps
 from LabExT.Movement.PathPlanning import PathPlanning
 
-from LabExT.Movement.config import State, Orientation, DevicePort
+from LabExT.Movement.config import CLOCKWISE_ORDERING, State, Orientation, DevicePort
 from LabExT.Movement.Calibration import Calibration
 from LabExT.Movement.Stage import Stage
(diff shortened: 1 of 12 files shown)


LabExT/Movement/config.py
```diff
@@ -41,6 +41,13 @@ class Orientation(BaseEnum):
     BOTTOM = auto()
 
 
+CLOCKWISE_ORDERING = [
+    Orientation.TOP,
+    Orientation.LEFT,
+    Orientation.BOTTOM,
+    Orientation.RIGHT]
```
marcoep, line 48: This is NOT clockwise, this is anti-clockwise! :P

LabExT/Movement/MoverNew.py
```diff
@@ -498,3 +570,23 @@ def _get_calibration(self, port=None, orientation=None, default=None):
             port)
         port = port or self._port_by_orientation.get(orientation)
         return self.calibrations.get((orientation, port), default)
+
+    def _wait_for_stages_to_stop(self, timeout: float = 2.0):
```
marcoep, line 574: Might be a very short timeout? Depending on the speed and distance, this time could be quite long?

marcoep (changes requested):

marcoep (approved):

PR #99, opened 2022-08-31
[Movement] Apply new Mover everywhere

maltewae:
🎉 
Applies the new mover everywhere (except for SfP, PR is coming soon).
Deletes the old mover.


LabExT/View/MainWindow/MainWindowModel.py
```diff
@@ -83,10 +83,10 @@ def __init__(self, controller, root, experiment_manager):
...
 
         # status of various sub-modules
-        self.status_mover_driver_enabled = BooleanVar(self.root)
-        self.status_mover_driver_enabled.trace("w", self.submodule_status_updated)
-        self.status_transformation_enabled = BooleanVar(self.root)
-        self.status_transformation_enabled.trace("w", self.submodule_status_updated)
+        self.status_mover_connected_stages = BooleanVar(self.root)
+        self.status_mover_connected_stages.trace("w", self.submodule_status_updated)
+        self.status_mover_fully_calibrated = BooleanVar(self.root)
+        self.status_mover_fully_calibrated.trace("w", self.submodule_status_updated)
```
marcoep, line 89: Careful with naming: `status_mover_fully_calibrated` is not really correct, no?

Suggest to rename this to `status_mover_can_move_to_device` or something similar. Because in the end, moving absolutely is enabled as soon as we have a single point fixed.

marcoep (approved):
well done!

LabExT/View/MainWindow/MainWindowModel.py
maltewae: Good point, I renamed it.

PR #100, opened 2022-09-19
[Movement]: Make SfP applicable for new mover

maltewae:
Makes `SearchForPeak` workable with new Mover without changing much in the general structure of the algorithm. When we redesign SfP, we can make this prettier. All fixes discovered during testing in the lab are included.

### DIFF
--- a/LabExT/Movement/Calibration.py
+++ b/LabExT/Movement/Calibration.py
@@ -14,7 +14,7 @@ from time import sleep
 
 import numpy as np
 
-from LabExT.Movement.config import DevicePort, Orientation, State, Axis, Direction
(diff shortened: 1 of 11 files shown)


marcoep (approved):
Looks good. I'm interested how it turns out when testing :)

maltewae:
@marcoep @boris-vukovic  @delmedico  Updates search for peak for new coordinate control. Only difference: I now use the new context manager `with self.mover.set_stages_coordinate_system(CoordinateSystem.STAGE)` to set the system of all stages to STAGE system. Before I did this manually.

marcoep (approved):

PR #101, opened 2022-11-02
[Movement] Update stage lifting/lowering

maltewae:
Updates the z lift and lifting behaviour of stages during automatic movement of stages to device.

The following is need:
- Updated Path Finding Algo: Now you can set a z level in chip coordinates for each stage. If given the the stage, will also keep the z coordinate at this level. Note that we keep the chip coordinate z level equal, not the stage coordinate z level. This coordinate will be transformed in stage space. This can incorporate a slight tilt of the chip. If no level is z, the algo always tries to keep the z coordinate the same as before.
- Enhanced the lifting logic: Now we keep track of lifted stages individually. The idea is that we only lift or lower stages, if the stage was lowered or lifted before.

### DIFF
--- a/LabExT/Movement/MoverNew.py
+++ b/LabExT/Movement/MoverNew.py
@@ -641,8 +641,11 @@ class MoverNew:
                                 atol=1e-08).all():
                             self.logger.error(
                                 f"{calibration} didn't not move properly to {target_z} while lifting/lowering stage.")
-
-                        self._lifted_stages.add(calibration)
+                        
+                        if _is_lifted:
+                            self._lifted_stages.remove(calibration)
+                        else:
+                            self._lifted_stages.add(calibration)
 
         _toggle_stage_lift()
         try:
--- a/LabExT/Movement/MoverNew.py
+++ b/LabExT/Movement/MoverNew.py
@@ -641,6 +641,7 @@ class MoverNew:
                                 atol=1e-08).all():
                             self.logger.error(
                                 f"{calibration} didn't not move properly to {target_z} while lifting/lowering stage.")
+                            return
                         
                         if _is_lifted:
                             self._lifted_stages.remove(calibration)
--- a/LabExT/Movement/MoverNew.py
+++ b/LabExT/Movement/MoverNew.py
@@ -239,27 +239,27 @@ class MoverNew:
             c.state >= State.COORDINATE_SYSTEM_FIXED for c in self.calibrations.values())
 
     @property
-    def left_calibration(self): return self._get_calibration(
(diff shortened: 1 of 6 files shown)


LabExT/Movement/MoverNew.py
```diff
@@ -602,17 +615,40 @@ def lift_stages(self):
...
+                    target_coordinate = ChipCoordinate(
+                        x=stage_start_pos.x, y=stage_start_pos.y, z=target_z)
+                    try:
+                        calibration.move_absolute(
+                            target_coordinate, wait_for_stopping=True)
+                    finally:
+                        stage_end_pos = calibration.get_position()
+                        lift_delta = stage_end_pos - stage_start_pos
+
+                        if not np.isclose(
```
maltewae, line 637: @marcoep @boris-vukovic Need same feedback on the tolerances.
```
absolute(a - b) <= (atol + rtol * absolute(b))
```
`atol` and `rtol` are the default ones at the moment.

LabExT/Movement/MoverNew.py
```diff
@@ -602,17 +615,44 @@ def lift_stages(self):
...
```
marcoep, line 637: Why are you doing this check here? Why could this fail?

LabExT/Movement/MoverNew.py
```diff
@@ -517,7 +525,12 @@ def move_absolute(
                 raise MoverError(
                     f"No {orientation} stage configured, but target coordinate for {orientation} passed.")
 
-            path_planning.set_stage_target(calibration, target)
+            current_stage_level = None
+            if calibration in self._lifted_stages:
+                current_stage_level = self.z_lift
```
marcoep, line 530: At this point your stages should be lifted anyhow.

I cannot see a reason why you would pass a z_level to the path planning. Doesn't path planning just move on a fixed z-level? If you start lifted, the path planning should stay on the lifted z-level anyhow, no?

LabExT/Movement/MoverNew.py
```diff
@@ -602,17 +615,40 @@ def lift_stages(self):
...
         """
 
-        def _lift_lower_stages(lift):
+        def _toggle_stage_lift():
+            """
+            Moves all stages to z lift level.
+            Moves up if stages_are_lifted is false and vice versa.
+            """
             for calibration in self.calibrations.values():
+                _is_lifted = calibration in self._lifted_stages
```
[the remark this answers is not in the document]
marcoep: Why do you error check the lifting movement? Is this prone to fail?

maltewae:
See #103

PR #103, opened 2022-11-09
[Movement v1] Lifting and Lowering of Stages

maltewae:
# td;lr
Lift and Lower is now the responsibility of the calibration. Path finding keeps the z-coordinate constant during movement.

# Detailed explanation
Changed the lift and lowering of the stages. Added methods to lift and lower stages at the calibration level (Recall: `Calibration` is an abstraction of a stage that holds all the information to move stages in chip or stage coordinate system, e.g., Kabsch rotation). Each calibration stores a bool `_is_lifted`, indicating whether the stage is currently lifted. This prevents double lifting or lowering.

## Logic to absolute lifting:
A `z-lift` value is passed. The stage moves in chip coordinates to `[x,y,z-lift],` where `x` and `y` are the current `x` and `y` values.

## The logic for absolute lowering:
The stage moves (independent of the previously passed `z-lift` value) to the position `[x,y,0]`, where `x` and `y` are the stage's current `x` and `y` values.

Both movements are carried out with `wait_for_stopping`.

## Changes in the mover:
The lift and lower methods have been removed in the `Mover`, and the logic has been simplified. The `move_to_device` method was adapted.

### The logic of the `move_to_device` method:
A check is made to see whether an input and output calibration has been defined. A lift is first ordered for these calibrations. Then both stages are moved absolutely. The wayfinding was adjusted to keep the Z position constant. This reads the start z-value of the stage and then keeps it stable. Alternatively, the requested z-value can be passed directly to prevent inaccuracies*. After the path-finding, the calibration is lowered.
Furthermore, the target coordinates are adjusted with this z-value. This fixes the last strange movement step we had in the lab.

*Why approximately: Assuming the stage is to be lifted to 50 before invoking the Pathfinding. Then the chip coordinate [x,y,50] is passed to Kabsch so that a stage coordinate [x',y',z'] is created. Uncertainty arises here in the translation. Then the stage is driven to the position, here uncertainty arises in the driving. Then comes the path-finding algorithm: the position of the stage is read out in stage coordinates and translated back into chip coordinates with Kabsch. This creates uncertainty in reading out and translating back.

@marcoep Happy to discuss the last point if it makes sense to pass the z-lift to the path finding to reduce uncertainty instead of reading it out.

### DIFF
--- a/LabExT/Movement/Calibration.py
+++ b/LabExT/Movement/Calibration.py
@@ -571,8 +571,52 @@ class Calibration:
             z=stage_coordinate.z,
             wait_for_stopping=wait_for_stopping)
 
+    def lift_stage(self, z_lift: float) -> None:
+        """
+        Lifts the stage absolutely or relatively.
+        Lifts stage absolutely if calibration is fully calibrated.
+        Lifts stage relatively if calibration is single point fixex or coordinate system fixed.
+        Parameters
+        ----------
+        z_lift: float
+            Amount in um to move the stage up.
+        Raises
+        ------
+        CalibrationError
+            If stage cannot lifted in the current state.
+        """
+        if self.state == State.FULLY_CALIBRATED:
+            self.lift_stage_absolute(z_lift)
+        elif self.state == State.SINGLE_POINT_FIXED or self.state == State.COORDINATE_SYSTEM_FIXED:
+            self.lift_stage_relative(z_lift)
+        else:
+            raise CalibrationError(
+                f"Cannot lift stage {self}: Calibration must be at least in coordinate fixed state.")
+
+    def lower_stage(self, z_lift: float) -> None:
+        """
+        Lowers the stage absolutely or relatively.
+        Lowers stage absolutely if calibration is fully calibrated.
+        Lowers stage relatively if calibration is single point fixex or coordinate system fixed.
+        Parameters
+        ----------
+        z_lift: float
+            Amount in um to move the stage down.
+        Raises
+        ------
+        CalibrationError
+            If stage cannot lowered in the current state.
+        """
+        if self.state == State.FULLY_CALIBRATED:
+            self.lower_stage_absolute()
+        elif self.state == State.SINGLE_POINT_FIXED or self.state == State.COORDINATE_SYSTEM_FIXED:
+            self.lower_stage_relative(z_lift)
+        else:
+            raise CalibrationError(
+                f"Cannot lower stage {self}: Calibration must be at least in coordinate fixed state.")
+
     @assert_minimum_state_for_coordinate_system(
-        chip_coordinate_system=State.SINGLE_POINT_FIXED)
+        chip_coordinate_system=State.FULLY_CALIBRATED)
     def lift_stage_absolute(self, z_lift: float) -> None:
         """
         Lifts the stage absolutely.
@@ -612,7 +656,39 @@ class Calibration:
             self._is_lifted = True
 
     @assert_minimum_state_for_coordinate_system(
-        chip_coordinate_system=State.SINGLE_POINT_FIXED)
+        chip_coordinate_system=State.COORDINATE_SYSTEM_FIXED)
+    def lift_stage_relative(self, z_lift: float) -> None:
+        """
+        Lifts the stage relatively.
+        Suppose the current position of the stage is [x_curr, y_curr, z_curr] in chip coordinates.
+        The stage will be moved relatively in chip coordinates to [x_curr, y_curr, z_curr + z_lift]
+        while z_lift is the input of the method.
+        Does not lift the stage if the stage is in a lifted state.
+        Parameters
+        ----------
+        z_lift: float
+            Amount in um to move the stage up.
+        """
+        if self.is_lifted:
+            self._logger.warn(
+                f"Stage {self} is already lifted. Skipping lift.")
+            return
+
+        self._logger.debug(
+            f"Lifting {self} relative up {z_lift} um")
+
+        with self.perform_in_system(CoordinateSystem.CHIP):
+            target_offset = ChipCoordinate(x=0, y=0, z=z_lift)
+            try:
+                self.move_relative(target_offset, wait_for_stopping=True)
+            except BaseException:
+                self._is_lifted = False
+                raise
+
+            self._is_lifted = True
+
+    @assert_minimum_state_for_coordinate_system(
+        chip_coordinate_system=State.FULLY_CALIBRATED)
     def lower_stage_absolute(self) -> None:
         """
         Lowers the stage absolutely.
@@ -639,6 +715,38 @@ class Calibration:
                 self.move_absolute(
                     target_lower_coordinate,
                     wait_for_stopping=True)
+            except BaseException:
+                self._is_lifted = True
+                raise
+
+            self._is_lifted = False
+
+    @assert_minimum_state_for_coordinate_system(
+        chip_coordinate_system=State.COORDINATE_SYSTEM_FIXED)
+    def lower_stage_relative(self, z_lift: float) -> None:
+        """
+        Lowers the stage relatively.
+        Suppose the current position of the stage is [x_curr, y_curr, z_curr] in chip coordinates.
+        The stage will be moved relatively in chip coordinates to [x_curr, y_curr, z_curr - z_lift]
+        while z_lift is the input of the method.
+        Does not lower the stage if the stage is in a lowered state.
+        Parameters
+        ----------
+        z_lift: float
+            Amount in um to move the stage down.
+        """
+        if not self.is_lifted:
+            self._logger.warn(
+                f"Stage {self} is already lowered. Skipping lowering.")
+            return
+
+        self._logger.debug(
+            f"Lowering {self} relative by {z_lift} um")
+
+        with self.perform_in_system(CoordinateSystem.CHIP):
+            target_offset = ChipCoordinate(x=0, y=0, z=-z_lift)
+            try:
+                self.move_relative(target_offset, wait_for_stopping=True)
             except BaseException:
                 self._is_lifted = True
                 raise
--- a/LabExT/Movement/Calibration.py
+++ b/LabExT/Movement/Calibration.py
@@ -51,15 +51,25 @@ def assert_minimum_state_for_coordinate_system(
                     "Function {} needs a cooridnate system to operate in. Please use the context to set the system.".format(
                         func.__name__))
 
-            if calibration.coordinate_system == CoordinateSystem.CHIP and calibration.state < chip_coordinate_system:
-                raise CalibrationError(
-                    "Function {} needs at least a calibration state of {} to operate in chip coordinate system".format(
-                        func.__name__, chip_coordinate_system))
+            if calibration.coordinate_system == CoordinateSystem.CHIP:
+                if chip_coordinate_system is None:
+                    raise CalibrationError(
+                        "Function {} does not support the chip coordinate system.".format(func.__name__))
+                
+                if calibration.state < chip_coordinate_system:
+                    raise CalibrationError(
+                        "Function {} needs at least a calibration state of {} to operate in chip coordinate system".format(
+                            func.__name__, chip_coordinate_system))
 
-            if calibration.coordinate_system == CoordinateSystem.STAGE and calibration.state < stage_coordinate_system:
-                raise CalibrationError(
-                    "Function {} needs at least a calibration state of {} to operate in stage coordinate system".format(
-                        func.__name__, stage_coordinate_system))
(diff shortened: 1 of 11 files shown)


LabExT/Movement/MoverNew.py
```diff
@@ -615,54 +621,52 @@ def move_relative(
...
-                ChipCoordinate(z=self.z_lift)
+        # Defines movement commands: Mapping from orientation to Chip
+        # coordinate
+        with self.set_stages_coordinate_system(CoordinateSystem.CHIP):
+            movement_commands = {}
+
+            if self.input_calibration:
+                # Input stage is defined.
+                # Lift stage and add input coordinate to movement commands
+                self.input_calibration.lift_stage_absolute(self.z_lift)
```
marcoep, line 648: Please move all lifting/lowering stages into `move_absolute`.

LabExT/Movement/MoverNew.py
```diff
@@ -615,54 +621,52 @@ def move_relative(
...
+                movement_commands[self.input_calibration.orientation] = device.input_coordinate
 
-        with self.lift_stages():
-            self.move_absolute(movement_commands, chip=chip)
+            if self.output_calibration:
+                # Output stage is defined.
+                # Lift stage and add output coordinate to movement commands
+                self.output_calibration.lift_stage_absolute(self.z_lift)
```
marcoep, line 654: Please move all lifting/lowering stages into `move_absolute`.

LabExT/Movement/MoverNew.py
```diff
@@ -615,54 +621,52 @@ def move_relative(
...
+                    chip=chip,
+                    stages_are_lifted=True)
+            finally:
+                if self.input_calibration:
+                    # Lower input stage
+                    self.input_calibration.lower_stage_absolute()
+
+                if self.output_calibration:
+                    # Lower output stage
+                    self.output_calibration.lower_stage_absolute()
```
marcoep, line 669: Please move all lifting/lowering stages into `move_absolute`.

LabExT/Movement/MoverNew.py
```diff
@@ -504,7 +504,8 @@ def move_absolute(
         movement_commands: Dict[Orientation, Type[ChipCoordinate]],
         chip: Type[Chip],
         wait_for_stopping: bool = True,
-        wait_timeout: float = 2.0
+        wait_timeout: float = 2.0,
+        stages_are_lifted: bool = False
```
marcoep, line 508: Replace this flag with a selector: `move_stages_lifted: bool = True`.

This flag should select if the stages should be lifted-moved-lowered or just moved.

LabExT/Movement/MoverNew.py
```diff
@@ -531,6 +535,7 @@ def move_absolute(
             return
 
         path_planning = PathPlanning(chip)
+        path_finding_z_lift = self.z_lift if stages_are_lifted else None
```
marcoep, line 538: Move stage lifting logic from `move_to_device` to here.

LabExT/Movement/PathPlanning.py
```diff
@@ -186,7 +186,8 @@ class PotentialField:
     def __init__(
         self,
         calibration: Type["Calibration"],
-        target_position: Type[ChipCoordinate],
+        target_coordinate: Type[ChipCoordinate],
+        start_z_lift: float = None,
```
marcoep, line 190: Get rid of this parameter. `PathPlanning` should always just stay on the chip z-coordinate of each starting point, i.e. should always move on the fixed xy-plane on a fixed z-level.

All the lifting/lowering should be done elsewhere, i.e. `move_absolute`.

LabExT/Movement/PathPlanning.py
```diff
@@ -358,7 +378,8 @@ def __init__(self, chip) -> None:
     def set_stage_target(
             self,
             calibration: Type["Calibration"],
-            target: Type[ChipCoordinate]) -> None:
+            target: Type[ChipCoordinate],
+            z_lift: float = None) -> None:
```
marcoep, line 382: Do we still need this param? `PathPlanning` should not know anything about lifting/lowering.

LabExT/Movement/Calibration.py
```diff
@@ -552,6 +571,80 @@ def move_absolute(self,
             z=stage_coordinate.z,
             wait_for_stopping=wait_for_stopping)
 
+    @assert_minimum_state_for_coordinate_system(
+        chip_coordinate_system=State.SINGLE_POINT_FIXED)
```
marcoep, line 575: Do I understand this correctly that this fct disallows using `lift_stage_absolute` in stage coordinates?

marcoep (changes requested):
I like having lifting/lowering purely in `Calibration`, this allows nice, concise checking for calib state etc.

However, I don't like that `move_to_device` and `PathPlanning` are involved in the lifting/lowering stages. In my understanding:
* `move_top_device` should only be concerned for calling `move_absolute` with the correct target coordinates, and nothing more.
* `PathPlanning` should only return waypoints on a z-level-fixed, 2D plane in (3D) chip coordinates. 

Now the lowering/lifting logic and checks are spread across at least 3 functions, which makes it hard to get correct and understand.

The purpose of `move_absolute` should be to move the stages to a target coordinate. With parameters, the user can choose:
* if collision-avoidance algorithm should be used, or simply moved directly (feature isn't implemented now, can omit for now)
* if all involved stages should move at the same time or sequentially
* if the stages should be z-level-lifted for the movement

Let me know if you have more questions.

LabExT/Movement/MoverNew.py
```diff
@@ -530,36 +533,49 @@ def move_absolute(
         if not movement_commands:
             return
 
-        path_planning = PathPlanning(chip)
+        with self.set_stages_coordinate_system(CoordinateSystem.CHIP):
+            path_planning = PathPlanning(chip)
```
maltewae, line 537: Changed the logic of `move_absolute`. A flag `with_lifted_stages` indicates whether the stages should be lifted/lowered beforehand. Path Planning knows nothing about this, it now simply reads the start Z-coordinate in the chip system and keeps it constant during the movement.

LabExT/Movement/MoverNew.py
```diff
@@ -530,36 +533,49 @@ def move_absolute(
...
+            resolved_calibrations = {}
+            for orientation, target in movement_commands.items():
+                calibration = self._get_calibration(orientation=orientation)
+                if calibration is None:
+                    raise MoverError(
+                        f"No {orientation} stage configured, but target coordinate for {orientation} passed.")
 
-            path_planning.set_stage_target(calibration, target)
+                if with_lifted_stages:
+                    calibration.lift_stage_absolute(self.z_lift)
```
maltewae, line 550: Here the stages are lifted.

LabExT/Movement/MoverNew.py
```diff
@@ -530,36 +533,49 @@ def move_absolute(
...
+                            break
 
-                    if all(c.stage.is_stopped
-                           for c in calibration_waypoints.keys()):
-                        break
+            # Movement complete and lower stages (if requested)
+            # Lift stages if requested
+            if with_lifted_stages:
+                for c in resolved_calibrations.values():
+                    c.lower_stage_absolute()
```
maltewae, line 578: Here the stages are lowered again.

LabExT/Movement/MoverNew.py
```diff
@@ -615,54 +631,37 @@ def move_relative(
...
+        if input_orientation:
+            movement_commands[input_orientation] = device.input_coordinate
+
+        if output_orientation:
+            movement_commands[output_orientation] = device.output_coordinate
+
+        self.move_absolute(
+            movement_commands,
+            chip=chip,
+            with_lifted_stages=True)
```
maltewae, line 664: `move_to_device` is now very simple: It looks at which orientation an input or output stage was configured. Accordingly, the movement commands (mapping from orientation to target chip coordinate) are constructed and passed to the move absolute.

LabExT/Movement/PathPlanning.py
```diff
@@ -230,23 +238,28 @@ def __init__(
...
+                np.argmin(np.abs(self.y_coords - self.start_coordinate.y))])
 
-        with self.calibration.perform_in_system(CoordinateSystem.CHIP):
-            self.start_coordinate = self.calibration.get_position()
-            self.target_position.z = self.start_coordinate.z
+        # current position in terms of field tiles
+        self.current_position = ChipCoordinate(
+            x=self.x_coords[self.current_idx[0]],
+            y=self.y_coords[self.current_idx[1]],
+            z=self.stage_z_level)
```
maltewae, line 250: Keep z level constant here.

LabExT/Movement/PathPlanning.py
```diff
@@ -230,23 +238,28 @@ def __init__(
...
 
-            self.current_idx = np.array(
-                [np.argmin(np.abs(self.x_coords - self.start_coordinate.x)),
-                 np.argmin(np.abs(self.y_coords - self.start_coordinate.y))])
+        # field target, note: z-coordinate remains unchanged
+        self.target_coordinate = ChipCoordinate(
+            x=target_coordinate.x,
+            y=target_coordinate.y,
+            z=self.stage_z_level)
```
maltewae, line 256: Keep z level constant here.

LabExT/Movement/PathPlanning.py
```diff
@@ -259,18 +272,18 @@ def next_waypoint(self) -> list:
...
-            self.current_position = self.target_position
+            self.current_position = self.target_coordinate
         else:
             self.current_idx = self._find_lowest_potential()
 
             self.current_position = ChipCoordinate(
                 x=self.x_coords[self.current_idx[0]],
                 y=self.y_coords[self.current_idx[1]],
-                z=self.start_coordinate.z)
+                z=self.stage_z_level)
```
maltewae, line 286: Keep z level constant here.

maltewae:
I understand your changes and have implemented them. 😄  I agree that path planning should have nothing to do with lowering and lifting. The only reason I passed the variable was to counteract the possible inaccuracies explained above, but they may not make a difference in the end.

LabExT/Movement/Calibration.py
maltewae: Yep, because we need the position in chip coordinates. 
I am working on a `lift_stage_relative` method (PR is almost ready) that moves the stages up and down by just adding an offset. Here we need at least the state "Coordinate System fixed"
The idea was that we lift fixed relative for a single point (i.e., take the `lift_stage_relative` method) and take the `lift_stage_absolute` method from Kabsch (since the tilting is then in).

maltewae: #104

LabExT/Movement/PathPlanning.py
```diff
@@ -210,14 +210,22 @@ def __init__(
...
 
+        # Fields settings
         self.grid_size = grid_size
         self.grid_outline = grid_outline
         self.repulsive_gain = repulsive_gain
         self.attractive_gain = attractive_gain
 
+        # Read current stage position
+        with self.calibration.perform_in_system(CoordinateSystem.CHIP):
+            self.start_coordinate = self.calibration.get_position()
```
marcoep, line 224: I suggest you add an error check. Since we know at which z-level we start (`self.stage_z_level`), simply check this value to be somewhat close to the target coordinates z-level (`target_coordinate.z`). If it's more than 1um away, raise an error.

maltewae: Okay, but for this I need to pass the z-lift to the PathPlanning. like I had before. 😅

marcoep: Umm, why? You have both variables `self.stage_z_level` and `target_coordinate.z` available in your local scope. No need for any external flag.

maltewae: `target_coordinate.z` is always 0 unless I change the z-value before passing it to the planner.
And `self.stage_z_level` is always going to be `start_coordinate.z` unless I pass more information to the planner. So I cannot really implement the check on the Planner level (without passing more information to it)

Or do I miss something here?

marcoep: Sorry for the delay.

No, you are fine. Still the check makes sense in the planner to make sure that start an end z-heights are equal. Because this is a fundamental assumption in your planner.

It does not matter if this check always passes with the current implementation of mover. Generally, adding checks to see if the inputs to an algorithm adhere to the assumptions that said algorithm works with is always (IMO) a good idea, no?

marcoep (approved):

LabExT/Movement/PathPlanning.py
maltewae: I agree! Opened #114 for this and merge this PR for now 😄

PR #105, opened 2022-11-16
[Movement v1] Fix bug of reset axes rotation after declining calibration loading

maltewae:
Fixes bug of resetting axes rotation after declining calibration loading.

Before we created a new `AxesRotation` after declining which was not updated during the calibration process such that Kabsch did not correct the unit vectors during calibration.

### DIFF
--- a/LabExT/Movement/MoverNew.py
+++ b/LabExT/Movement/MoverNew.py
@@ -6,12 +6,13 @@ This program is free software and comes with ABSOLUTELY NO WARRANTY; for details
 """
 
 import json
+import os
+import logging
 
 from time import sleep, time
 from bidict import bidict, ValueDuplicationError, KeyDuplicationError, OnDup, RAISE
 from typing import Dict, Tuple, Type, List
 from functools import wraps
-from os.path import exists
 from datetime import datetime
 from contextlib import contextmanager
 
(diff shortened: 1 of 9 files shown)


marcoep (changes requested):
Wonderful! Would it make sense to add a unittest which specifically provokes the bug? Obv. in the end, it should pass, but I'm missing a piece of code which tells me what was wrong...

maltewae:
I would add integration tests in another PR. Then I can test the entire workflow.

The mistake was the following (which was really stupid):
All transformations (Axes Rotation, Single Point and Kabsch) have an `initialize` method. This is used to set up the transformation once and also to reset the transformation. In the calibrations (which manage these transformations) there are the methods `reset_single_point_offset` and `reset_kabsch_rotation`. These reset (as the name suggests) the transformations by calling the `initialize` method and then updating the status of the calibration. These methods are mainly used when a new chip is imported.

Here is the error:
The `initialize` method of the Kabsch rotation previously allowed `axes_rotation: Type[AxesRotation] = None`. If `axes_rotation` was None, a new `AxesRotation` object is created from which the calibration was not informed. Therefore, if the `AxesRotation` of the calibration (other `AxesRotation` object than the one in the `KabschRotation`) was changed, the `KabschRotation` was not informed. This meant that the unit vectors could not be adjusted. This resulted in the going down instead of the going up when lifting.

marcoep (approved):

marcoep:
Thank you for the explanation. This is good to merge.

PR #106, opened 2022-11-20
[Movement v2]: Path Planning for single stage

maltewae:


boris-vukovic (changes requested):
Function looks correct, but the nomenclature in the docstring is a bit unclear to me. I would call the normal vector n and ux, uy, uz the unit vectors or something similar.

maltewae:
Closing this for #116 #117 #118 #119

PR #110, opened 2022-11-30
[Movement v1] SfP and Live Viewer shortcuts

maltewae:
This PR adds two optional shortcuts in the Coordinate Pairing Window for Search for Peak and Live Viewer. This makes pairing creation easier.

Screenshot 2022-11-30 at 14 33 46

In the future, when we define the new interface for SfP, we can directly set the required stages to perform the SfP Algorithm.

### DIFF
--- a/LabExT/View/Movement/CoordinatePairingsWindow.py
+++ b/LabExT/View/Movement/CoordinatePairingsWindow.py
@@ -7,13 +7,10 @@ This program is free software and comes with ABSOLUTELY NO WARRANTY; for details
 
 from tkinter import Frame, Toplevel, Button, Label, messagebox, LEFT, RIGHT, TOP, X, BOTH, DISABLED, FLAT, NORMAL, Y
 from typing import Callable, Type
-from LabExT.Utils import run_with_wait_window, try_to_lift_window
+from LabExT.Utils import run_with_wait_window
 from LabExT.View.Controls.CustomFrame import CustomFrame
 from LabExT.View.Controls.DeviceTable import DeviceTable
 from LabExT.View.Controls.CoordinateWidget import CoordinateWidget, StagePositionWidget
-
-from LabExT.View.LiveViewer.LiveViewerController import LiveViewerController
-from LabExT.View.SearchForPeakPlotsWindow import SearchForPeakPlotsWindow
 
 from LabExT.Movement.Transformations import CoordinatePairing
 from LabExT.Movement.MoverNew import MoverNew
@@ -71,9 +68,6 @@ class CoordinatePairingsWindow(Toplevel):
         self.on_finish = on_finish
         self.with_input_stage = with_input_stage
         self.with_output_stage = with_output_stage
-
-        self._live_viewer_toplevel = None
-        self._search_for_peak_toplevel = None
 
         if self.chip is None:
             raise ValueError("Cannot create pairing without chip imported. ")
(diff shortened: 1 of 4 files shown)


LabExT/View/Movement/CoordinatePairingsWindow.py
```diff
@@ -161,6 +172,24 @@ def __setup__(self) -> None:
...
                 side=TOP, fill=X)
 
+        if self.experiment_manager:
+            shortcuts_frame = CustomFrame(self._main_frame)
+            shortcuts_frame.pack(side=TOP, fill=X, pady=5)
+
+            search_for_peak_button = Button(
+                shortcuts_frame,
+                text="Perform Search for Peak...",
+                command=self._on_perform_search_for_peak)
```
marcoep, line 182: Suggestion: I would use the function `self.experiment_manager.main_window.open_live_viewer` here, so you can re-use all the top-level checking logic from the main window.

LabExT/View/Movement/CoordinatePairingsWindow.py
```diff
@@ -161,6 +172,24 @@ def __setup__(self) -> None:
...
+            search_for_peak_button.pack(
+                side=RIGHT, fill=Y, pady=5, expand=0)
+
+            live_viewer_button = Button(
+                shortcuts_frame,
+                text="Open Live Viewer...",
+                command=self._on_open_live_viewer)
```
marcoep, line 189: Same here, reuse the function `self.experiment_manager.main_window.open_peak_searcher` with no need to re-code logic in this class.

LabExT/View/Movement/CoordinatePairingsWindow.py
```diff
@@ -252,7 +281,36 @@ def _ask_user_move_to_device(self):
...
+        if not self.experiment_manager:
+            return
+
+        if try_to_lift_window(self._search_for_peak_toplevel):
+            return
+
+        sfpp = SearchForPeakPlotsWindow(
+            parent=self,
+            experiment_manager=self.experiment_manager)
+        self._search_for_peak_toplevel = sfpp.plot_window
```
marcoep, line 313: same logic as in `LabExT/View/MenuListener.py`, no need to re-code

marcoep (changes requested):

maltewae:
Neat! Didn't saw the methods on main window level. Changed it 😄

LabExT/View/Movement/CoordinatePairingsWindow.py
```diff
@@ -161,6 +166,24 @@ def __setup__(self) -> None:
...
                 side=TOP, fill=X)
 
+        if self.experiment_manager:
+            shortcuts_frame = CustomFrame(self._main_frame)
+            shortcuts_frame.pack(side=TOP, fill=X, pady=5)
+
+            search_for_peak_button = Button(
+                shortcuts_frame,
+                text="Perform Search for Peak...",
+                command=self.experiment_manager.main_window.open_peak_searcher())
```
marcoep, line 176: I think you only need to pass the function, not the function-call's-return here. i.e. delete parentheses at end.

LabExT/View/Movement/CoordinatePairingsWindow.py
```diff
@@ -161,6 +166,24 @@ def __setup__(self) -> None:
...
+            search_for_peak_button.pack(
+                side=RIGHT, fill=Y, pady=5, expand=0)
+
+            live_viewer_button = Button(
+                shortcuts_frame,
+                text="Open Live Viewer...",
+                command=self.experiment_manager.main_window.open_live_viewer())
```
marcoep, line 183: I think you only need to pass the function, not the function-call's-return here. i.e. delete parentheses at end.

marcoep (changes requested):

LabExT/View/Movement/CoordinatePairingsWindow.py
maltewae: Yes! 😅 I was too fast.

marcoep (approved):

PR #114, opened 2022-12-01
[Movement v1] Path Planning z level check

maltewae:
Adds a check if target and start coordinates are approximately on the same z plane.

### DIFF
--- a/LabExT/Movement/PathPlanning.py
+++ b/LabExT/Movement/PathPlanning.py
@@ -254,7 +254,7 @@ class PotentialField:
                 self.start_coordinate.z,
                 self.target_coordinate.z,
                 rtol=1.e-5,
-                atol=1.e-8):
+                atol=10e-3):
             raise ValueError(
                 f"Start z level {self.start_coordinate.z} is not close to target z level {self.target_coordinate.z}. "
                 "The Path Planning algorithm assumes that start and target are on the same z level.")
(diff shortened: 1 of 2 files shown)


LabExT/Movement/PathPlanning.py
```diff
@@ -247,11 +247,17 @@ def __init__(
...
-            y=target_coordinate.y,
-            z=self.start_coordinate.z)
+        # field target
+        self.target_coordinate = target_coordinate
+
+        if not np.isclose(
+                self.start_coordinate.z,
+                self.target_coordinate.z,
+                rtol=1.e-5,
+                atol=1.e-8,):
```
maltewae, line 257: I took the default values here. Are they ok, @marcoep ?

maltewae: Don't wonder, I just removed the comma after `atol=1.e-8,`

marcoep: I would set it such that it is okay within a +-10nm range. So keep `rtol` the same, but change `atol` to something like 10e-2 (assuming we are in units of microns).

maltewae: Changed it to 10e-2. Yes all units are in um.

marcoep: ahh, sorry, meant 10e-3 (i.e. 10nm)

maltewae: okay its 10e-3 now 😄

PR #115, opened 2022-12-01
🚂🚀 New LabExT Movement implementation

maltewae:
# LabExT Movement reimplementation 

This PR summarises all changes in the Movement in LabExT. The changes are briefly described below:

## Stages
The Stages class defines an interface for all Stages that LabExT is to support. As soon as an implementation of a Stage inherits from the `Stage` base class, it is automatically found in LabExT and can be registered in the Mover for further use.

Currently, there is a stage implementation for SmarAct and a simulation stage.

## Calibrations
A calibration provides abstractions for the safe use of the stages. A calibration takes a stage and defines and stores all information to allow the given stage to be moved in any coordinate system. In LabExT there are two coordinate systems: stage and chip. The context manager `perform_in_system` can be used to define in which coordinate system the stage should operate.

Operations in chip system:
```python
with calibration.perform_in_system(CoordinateSystem.CHIP):
   pass
```

Operations in stage system:
```python
with calibration.perform_in_system(CoordinateSystem.STAGE):
   pass
```

At the calibration level, the stage can be moved absolutely with `move_absolute` and relatively with `move_relative`. Both functions can be used in the stage and chip system. If the stage system is requested, the stage is addressed directly without changes. If the chip system is to be used, various transformations are used.

### Transformations
The Movement Update defines 3 transformations:
- **Axis rotation:** The axis rotation defines a 3x3 signed permutation matrix. It can be used to define which stage axis is responsible for which chip axis. The goal is, no matter where the stage is located in space, that all stages are aligned with the chip axis. What is new compared to before: Currently only the Z-axis can be inverted. This is still possible, except that you can now reassign all axes. This rotation is saved on the PC and can be restored.
- **Single-point offset:** In order to be able to move absolute in chip coordinates more quickly, we define an offset transfromation. With the help of a point pairing (pairing of a chip and stage coordinate) and the axis rotation, an offset can be calculated. The stage can now move absolutely in chip coordianates, with the big assumption that the axes of the stage and the chip are parallel. Therefore, this is only an approximation to be able to move Stage absolute earlier.
- **KabschRotation:** From 3 point pairings (new: more are possible, from more than 3 the Kabsch is optimised with Least Squares) we define a full calibration that takes into account the inclination of the chip in all directions..

These transformations are used when requested in the chip system:
- `move_relative`: Axis rotation specifies the correct direction of the offset.
- `move_absolute`: Single point offset or Kabsch to translate the coordinate. If defined, Kabsch is always preferred.
- `get_position`: Single-point-offset or Kabsch to translate the stage coordinate into a chip coordinate. 

## Mover
The Mover is the entry point for movements in LabExT. It stores all registered calibrations and defines how they are in space (`LEFT`, `RIGHT`, `TOP`, `BOTTOM`) and whether they are to be used for the input or output of a device.

In the Mover, the coordinate system of all stages can be set directly.

All stages can be moved relative and absolute in the Mover.

### PathPlanning
The new mover defines a path-planning algorithm that works with the help of a potential field.

To do this, the chip is first divided into a chess field. The size of a tile is at most as large as the smallest distance between two device points on the chip. Each operating stage now defines its own potential field, with the target defining a positive potential and all other stages (aka obstacles) creating a negative potential. Each stage now makes an optimisation step to a next tile to gain potential. After that, the field is recalculated for all as the obstacles (aka stages) have moved. This is repeated until each Stage has reached the goal or each Stage has made no progress in the last 3 steps. In the latter, an error is thrown.

Each stage must define a simple polygon to participate in the algorithm. This polygon defines the obstacle's size and which tiles should be given a negative potential.

The algorithm compensates for the tilt of the chip.


marcoep (approved):
Given successful testing this afternoon, this MR is good to merge. Most additions were reviewed in previous 20 MRs or so, so this complete rewrite of the movement part is amazing!

maltewae:
@marcoep @boris-vukovic @delmedico Final approval to merge into dev 😄

delmedico (approved):
this is excellent! big thanks

boris-vukovic (approved):

PR #116, opened 2022-12-01
[Movement v2] Path Planning Abstraction

maltewae:
Introduction of an interface for path planning in the Mover.

## The interface is defined by:
- `set_stage_target`: Given a calibration instance and a chip coordinate as a target, we define which calibration has which target.
- `trajectory`: Generator that iteratively creates a waypoint command. Why generator: Generators are more general than simple lists, and allow the calculation of waypoints that are not yet known at the time of the call. The `CollisionAvoidancePlanning` with potential field (previously `PathPlanning`) uses this. 

Furthermore, `PathPlanning` has been renamed `CollisionAvoidancePlanning`. The algorithm is the same.

## Introduction of a waypoint tuple:
The `Waypoint` NamedTuple defines the calibration, coordinate, and `wait_for_stopping` for a calculated waypoint. Not really a big change than before, except that we get more unity within the class. With `wait_for_stopping` we can tell the stages that they should wait for this waypoint to be executed.

## Introduction of a type alias:
`WaypointCommand` is a type alias for a dict from `Calibration` to `Waypoint`. This dict is to be created by the generator in each step and specifies which calibration is to be run to which waypoint.
The introduction should provide unity.

### DIFF
--- a/LabExT/Movement/PathPlanning.py
+++ b/LabExT/Movement/PathPlanning.py
@@ -5,10 +5,12 @@ LabExT  Copyright (C) 2022  ETH Zurich and Polariton Technologies AG
 This program is free software and comes with ABSOLUTELY NO WARRANTY; for details see LICENSE file.
 """
 
-from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, List, Tuple, Type
-
+import logging
 import numpy as np
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, List, Any, Dict, NamedTuple, Tuple, Type, Generator
+
 from scipy.spatial.distance import pdist
 
 from LabExT.Movement.Transformations import ChipCoordinate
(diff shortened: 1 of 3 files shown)


marcoep (approved):
Very well thought out! Congratz!

LabExT/Movement/PathPlanning.py
```diff
@@ -464,23 +516,21 @@ def _get_grid_properties(
...
         """
-        if len(last_moves) == 0:
+        if len(self.last_moves) == 0:
             return False
 
-        for last_move in last_moves:
+        for last_move in self.last_moves[-self.abort_local_minimum:]:
             for calibration, waypoint in last_move.items():
-                if new_move[calibration] != waypoint:
+                if next_command[calibration].coordinate != waypoint.coordinate:
```
maltewae, line 533: @marcoep We found a bug during testing. I forgot to compare the coordinates and not the waypoint. I have updated the routine.

marcoep (approved):

PR #125, opened 2022-12-05
requirements.txt: upgraded dependencies

marcoep:
Upgrade of dependencies.


marcoep:
Dep update unnecessary at this point. Deps already upgraded in #115 .

PR #128, opened 2022-12-06
Disable context menu options if not available

maltewae:
Disables options in Context Menu for Movement if they are not available in the current state.

### DIFF
--- a/LabExT/View/MainWindow/MainWindowView.py
+++ b/LabExT/View/MainWindow/MainWindowView.py
@@ -8,7 +8,8 @@ This program is free software and comes with ABSOLUTELY NO WARRANTY; for details
 import logging
 from platform import system
 from tkinter import Frame, Menu, Checkbutton, \
-    Label, StringVar, OptionMenu, LabelFrame, Button, scrolledtext, Entry
+    Label, StringVar, OptionMenu, LabelFrame, Button, scrolledtext, Entry, \
+    NORMAL, DISABLED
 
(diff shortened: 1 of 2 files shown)


LabExT/View/MainWindow/MainWindowView.py
```diff
@@ -62,21 +63,26 @@ def __init__(self, parent, menu_listener):
             command=self._menu_listener.client_setup_stages)
         self._movement.add_command(
             label="Configure Mover...",
-            command=self._menu_listener.client_setup_mover)
+            command=self._menu_listener.client_setup_mover,
+            state=NORMAL if self._menu_listener._experiment_manager.mover.has_connected_stages else DISABLED)
```
marcoep, line 67: Is this really true that one can only configure the mover if there are connected stages? The mover object is existing anyhow before any stages are connected, no?

I do see why we could simply disable configuring the mover for workflow reasons, to make it less confusing for people... So I have nothing against this check but I'm just curious.

marcoep (approved):

LabExT/View/MainWindow/MainWindowView.py
maltewae: The current implementation requires connected stages. The mover wizard invokes e.g the following method https://github.com/LabExT/LabExT/blob/dev/LabExT/Movement/MoverNew.py#L431. This sets the property directly in the stage, so we need at least one connected stage.

On startup, when the mover gets initialised, we load these property only internally.

PR #129, opened 2022-12-06
Show mover state in context menu

maltewae:
Adds information about the Mover state in the Context Menu.

Screenshot 2022-12-06 at 19 12 39
Screenshot 2022-12-06 at 19 12 45


LabExT/View/MainWindow/MainWindowView.py
```diff
@@ -57,6 +59,21 @@ def __init__(self, parent, menu_listener):
...
+        if self._mover.has_connected_stages:
+            self._movement.add_command(
+                label=f"{len(self._mover.calibrations)} connected stage(s):",
+                state=DISABLED)
+            for c in self._mover.calibrations.values():
+                self._movement.add_cascade(
+                    label=str(c),
+                    menu=self._make_calibration_menu(c))
+            self._movement.add_separator()
+
```
marcoep, line 76: Is the menu truly always reflecting the status of the calibrations? I.e. is `_make_calibration_menu` always called anew when the user is moving the mouse over the menu? Or how do you make sure the menu entries are updated?

marcoep (changes requested):

LabExT/View/MainWindow/MainWindowView.py
maltewae: Yes the menu gets updated on every state change. If the user changes the single point offset for example, the mover invokes the `update_main_model` method. This method refreshes the context menu: https://github.com/LabExT/LabExT/pull/129/files#diff-c2f02b3fc37853d919db74aeb42940599e80aed0beb405cdf26bdea933e275a5R173

marcoep (approved):

maltewae:
Rebased to `dev`

PR #132, opened 2022-12-07
Update workflow for stage calibration when Chip is not imported

maltewae:
Currently we do not allow to open the Stage Calibration Wizard if no chip has been imported.
In particular, this does not allow you to adjust the axes, which does not require a chip.

The changes introduce an extra frame in the coordinate pairing step that shows the currently imported chip, which gives more information. If the chip is missing, we provide a shortcut to import a chip. The wizard will then be reloaded. If no chip has been imported, no coordinate pairings can be created. The buttons are therefore deactivated.

Without imported chip:
Screenshot 2022-12-07 at 16 18 08

With imported chip:
Screenshot 2022-12-07 at 16 18 18

### DIFF
--- a/LabExT/View/Movement/MovementWizard.py
+++ b/LabExT/View/Movement/MovementWizard.py
@@ -221,7 +221,32 @@ class MoverWizard(Wizard):
 
 
 class CalibrationWizard(Wizard):
-    def __init__(self, master, mover, chip, experiment_manager=None):
+    """
+    Wizard to calibrate stages.
+    """
+
+    def __init__(
+        self,
+        master,
+        mover,
+        chip=None,
+        experiment_manager=None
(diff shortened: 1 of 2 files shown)


LabExT/View/Movement/MovementWizard.py
```diff
@@ -230,9 +255,6 @@ def __init__(self, master, mover, chip, experiment_manager=None):
             raise RuntimeError(
                 "Calibration not possible without connected stages.")
 
-        if self.chip is None:
-            raise RuntimeError("Calibration not possible without loaded chip.")
```
maltewae, line 234: Removing this check allows to open the wizard without chip

marcoep (approved):

PR #140, opened 2022-12-11
Remove old driver dialog and use new one

maltewae:
PR removes the old driver dialog and uses the new one.
Added a method to restart LabExT


LabExT/View/MenuListener.py
```diff
@@ -194,6 +194,12 @@ def client_quit(self):
         """
         sys.exit(0)
 
+    def client_restart(self):
+        """
+        Called when user wants to restart the applications.
+        """
+        os.execl(sys.executable, sys.executable, *sys.argv)
```
maltewae, line 201: Short explanation: `os.execl` replaces the current process (LabExT) with a new one constructed by the arguments. `sys.executable` gives the path to python interpreter and `sys.argv` are the current program argument. Short: It just replaces the current LabExT with a new one, thus forcing to restart LabExt.

marcoep (approved):

PR #147, opened 2023-02-03
Addon support for stages

maltewae:
This PR allows the loading of stage classes from an arbitrary path from the disk, similar to Instruments and Measurements.

### DIFF
--- a/LabExT/ExperimentManager.py
+++ b/LabExT/ExperimentManager.py
@@ -67,13 +67,15 @@ class ExperimentManager:
         self.addon_settings = None
         self.chip = chip
         self.mover = MoverNew(experiment_manager=self, chip=chip)
-        self.peak_searcher = PeakSearcher(None, self, mover=self.mover, parent=self.root)
+        self.peak_searcher = PeakSearcher(
+            None, self, mover=self.mover, parent=self.root)
         self.instrument_api = InstrumentAPI(self)
         self.docu = None
         self.live_viewer_cards = {}
         self.lvcards_import_stats = {}
 
-        # make sure instruments.config file is present, as its used in the resource manager
+        # make sure instruments.config file is present, as its used in the
+        # resource manager
         instruments_are_default = self.setup_instruments_config()
 
         # create global unique resource manager
@@ -97,10 +99,12 @@ class ExperimentManager:
             # the fact that the _tkinter module attempts to gain control of the main thread via a polling technique
             # when processing calls from other threads.
             # This is why we need to put all the setting up into a different thread (yayy)
-            # since we are working in a diffrent thread, we need a variable to signal the end of the process
+            # since we are working in a diffrent thread, we need a variable to
+            # signal the end of the process
             self.setup_done = False
             # here we set up the progress bar
-            self.pgb = ProgressBar(root, 'Welcome to LabExt\nWe are setting everything up for you!')
+            self.pgb = ProgressBar(
+                root, 'Welcome to LabExt\nWe are setting everything up for you!')
             # this is needed, since tk automatically opens a root window, which we do not want. The withdraw
             # command hides that window
             root.withdraw()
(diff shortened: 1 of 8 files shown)


LabExT/View/AddonSettingsDialog.py
```diff
@@ -181,6 +195,15 @@ def _get_loaded_lvcards_and_paths(self):
                 ret_list.append(('    ->', '', vi))
         return ret_list
 
+    def _get_loaded_stages_and_paths(self) -> list:
```
delmedico, line 198: Now the Stages path from the core (.../LabExT/Movement/Stages) is listed in the addons list as default. Can we get rid of this, like it is done for the instruments and measurements?

LabExT/ExperimentManager.py
```diff
@@ -204,6 +204,10 @@ def setup_runner(self):
         self.exp.import_measurement_classes()
         # then we load all Instruments
         self.instrument_api.load_all_instruments()
+        # then we load all Stages and Settings
```
delmedico, line 207: the loaded stages should then also be added to the plugins loaded logging (done in the `__init__` method here) so it can be seen in the main window logging. Right now they are listed under lvcards

LabExT/Movement/Stage.py
```diff
@@ -63,16 +63,40 @@ class Stage(ABC):
     meta = StageMeta("", False)
 
     @classmethod
-    def find_stage_classes(cls, subdir="Stages") -> list:
+    def find_stage_classes(
+        cls,
+        search_paths: list = [],
```
delmedico, line 68: Could this cause the issue that paths are re-inserted even if they are already there?

LabExT/View/Movement/MovementWizard.py
```diff
@@ -346,15 +346,15 @@ def build(self, frame: Type[CustomFrame]) -> None:
...
 
-        for stage_cls in self.mover.stage_classes:
+        for stage_name, stage_cls in self.mover.stage_classes.items():
             stage_driver_frame = Frame(frame)
             stage_driver_frame.pack(side=TOP, fill=X, pady=2)
 
             Label(
                 stage_driver_frame,
-                text=f"[{stage_cls.__name__}] {stage_cls.meta.description}"
+                text=f"[{stage_name}] {stage_cls.meta.description}"
```
delmedico, line 357: My stages loaded from an addon path are shown here (drivers are loaded as well), but they are not listed in the stage assignment step afterwards. I used the dummy stage from the core loaded from my addon path.

delmedico (changes requested):

LabExT/View/AddonSettingsDialog.py
maltewae: I removed it. Now it looks like this:
Screenshot 2023-02-27 at 12 27 09

LabExT/ExperimentManager.py
maltewae: Added it, now here: 
https://github.com/LabExT/LabExT/pull/147/files#diff-01b358552e5d1a58354dc2dde4fe850bbf15cb68e78b2f69c8fc18e6b92099e8R151

LabExT/Movement/Stage.py
maltewae: Should work now, I do the import in the new `import_stage_classes` method of the mover. It will be invoked once by the setup runner, similar to measurement etc

LabExT/View/Movement/MovementWizard.py
maltewae: Yes, there was an error with the stage discovery. I changed it now, thanks! Can you check again if it works now?

marcoep:
@delmedico @maltewae where are we with this PR? Does Nino need to verify again?

maltewae:
@marcoep Rebased! 
I think Nino and I resolved comments in the meeting last Monday. Just need an approval.

marcoep (approved):

marcoep:
Should @delmedico have a look at it as well?

delmedico (approved):
I have tested this myself again. Well done and thanks!

PR #148, opened 2023-02-12
Prefill calibrations when restoring and make window to restore simpler

maltewae:
Automatically fills in the stage with which the calibration was created. This makes the restoring process of calibrations easier and safer. In addition, the code has been simplified.


marcoep (approved):
very well done :)

PR #149, opened 2023-02-13
Enable Swept Parameters

marcoep:
# Goal
Enable automated generation of multiple ToDos for sweeping parameters. No need for Meta-Measurements!

# Approach
Code was added that within the EditMeasurementWizard, new GUI elements allow setting-up any nummerical parameter for sweeping. Each parameter point then creates its own ToDo. The measurement records 
given by this ToDo include an additional field in the output data which includes information about the swept parameter settings.

# Implementation
tbd

# Compatibility
Code should not raise any compatibility issues.


marcoep:
Superseeded by #173

PR #150, opened 2023-02-14
Quality-of-life improvements for Keysight Laser / OPM

marcoep:
# Goal
Small improvements regarding the supplied Keysight TLS and OPM classes.

# Compatibility
Actually increases compatibility! N7744A class is actually compatible with all N774x keysight multichannel OPMs. The LaserMainframeKeysight class is compatible with all 8164 and 7776/7778 lasers.


delmedico (approved):
thank you!

PR #152, opened 2023-02-28
more robust TKinter unittests: separate tox environments for the GUI integration tests

marcoep:
After lots of headache, I have discovered that tkinter is quite garbage when we do the full integration tests of the Experiment Wizard and the Main Window.

I have also discovered that the integration tests do not provoke any tkinter race condition / multi-threading bugs. So we simply run them in their own environment. A small increase in per-test runtime for a huge decrease in headaches is IMHO worth it.

### DIFF
--- a/LabExT/Tests/Utils.py
+++ b/LabExT/Tests/Utils.py
@@ -20,6 +20,7 @@ import string
 RUNTESTS_DIR = path.abspath(path.dirname(__file__))
 TRANSFORMATIONS_DIR = path.join(RUNTESTS_DIR, "transformations")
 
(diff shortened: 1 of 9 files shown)


marcoep:
For some reason the tkinter tests now produce a segfault...

marcoep:
@maltewae @delmedico @boris-vukovic 

This is just a small QOL improvement again for the unittesting. Please review soon, it should not change much.

maltewae (approved):
Perfect, looks good 👍🏼 I like that we can disable these tests now.

PR #159, opened 2023-03-24
(2/8) Add orientation to Single Mode Fiber stage polygon

maltewae:
This pull request adds the orientation property as a StagePolygon property. Furthermore, StagePolygon is automatically saved and loaded on disk for class names by MoverAPI.


LabExT/View/Movement/MovementWizard.py
maltewae, line 1: This file will be completely revised in a later PR #163. Changes made here are then not always up to date. Therefore no focus here.

boris-vukovic (approved):

PR #168, opened 2023-07-17
InsertionLossSweep - add functionality to reference previous spectrum

marcoep:
# Goal
Make it possible to have a calibrated insertion loss plot created solely from LabExT without any further tool.

# Approach
The `InsertionLossSweep` measurement class has been extended such that it can read in a file path to a previously recorded spectrum and use the loaded data as a 0-loss-calibration. To successfully apply the calibration, the current measurement parameters MUST match the measurement parameters of the previous recording.

If a matching calibration file is supplied, there is a third output data vector added to the recorded data called 'referenced transmission [dB]' which is the recorded optical power ('transmission [dBm]') normalized to the loaded spectrum data.

# Implementation
If a calibration file is loaded, for simplicity we enforce that the measurement parameters must match the measurement parameters of the loaded calibration file.

A controversial point could be that we only add the third output data vector if a calibration file is loaded. This results in the case that sometimes a InsertionLossSweep dataset has two, sometimes three output data vectors...

# Compatibility
No break of compatibility, simple adding of feature.


LabExT/Measurements/InsertionLossSweep.py
```diff
@@ -181,7 +198,57 @@ def algorithm(self, device, data, instruments, parameters):
...
+        rec_wl = np.array(data['values']['wavelength [nm]'])
+        rec_tm = np.array(data['values']['transmission [dBm]'])
+
+        # notify user if wavelengths differ > 1pm
+        if any(np.abs(rec_wl - self.ref_data['wl']) > 1e-3):
+            self.logger.warning('Referenced and recorded wavelengths differ by > 1pm!')
+
+        # apply referencing
+        ref_data_diff = rec_tm - self.ref_data['tm']
+        data['values']['referenced transmission [dB]'] = ref_data_diff.tolist()
```
delmedico, line 251: Two things that I thought about here:
1. What do you think about the other way around; overwriting data['values']['transmission [dBm]'] with the calibrated transmission (ref_data_diff) and instead storing the "calibration transmission" as an additional value array. Most measurement tools that I know allow you to do a calibration but the calibrated output data does not change it's name. Additionally, all analysis tools and scripts that are based on these value keys would read the calibrated transmission as a default then.
2. What do you think about either making the storage of the calibration / uncalibrated data optional or not storing it all. Is referencing the measurement name / filepath enough? storing this data means approx. 50% of added file storage space. When executing a large number of measurements, this can easily become noticeable.

delmedico (commented):
super cool PR. thanks a lot!

LabExT/Measurements/InsertionLossSweep.py
marcoep: Thanks for the comment. Yes, I am not happy with only conditionally adding the extra output data vector either. I really like your approach 1.

The proposed discarding of the raw measurement data in your option two is risky for the longevity of the data, in my opinion. To make sense, one would have to remember to also save the calibration file that was used to take the measurement with - sometimes a challenge in chaotic research environments.

Based on your feedback, what do you, @boris-vukovic and @delmedico, think about the following changes to this PR:
* when loading a calibration during measurement, store the calibrated transmission in the vector `data['values']['transmission [dBm]']`
  * note that the unit of the referenced transmission is dB, not dBm - so full compatibility with the evaluation script would not be possible anyway
* in case calibration is applied during measurement, make storing the raw power meter values in a separate vector optional (additional boolean parameter)
* reworking this additional functionality into a separate measurement, i.e. not in `InsertionLossSweep`

boris-vukovic: Decisions taken 11.09.2023: @marcoep 
- Don't save different data (calibrated) under the same name of "transmission [dB]". Reason: Confuse people
- Add optional flag "only save calibrated data". Reason: Save disk space
- No new measurement class --> Implement in OG InsertionLossSweep

marcoep: done!

marcoep: @delmedico @boris-vukovic 

Check if you like the new implementation. I added another boolean parameter for the user to selectively discard the raw measurement data.

To note is that all save files now have the three data vectors (`wavelength [nm]`, `transmission [dBm]`, `referenced transmission [dB]`) always present as keys in the `data['values']` dict. If you did not chose to reference the data, the data is an empty list, similarly if you chose to discard the raw data.

Check if you like that solution and I'd be happy for some feedback.

LabExT/Measurements/InsertionLossSweep.py
```diff
@@ -181,7 +207,63 @@ def algorithm(self, device, data, instruments, parameters):
...
 
+        # apply reference to data
+        if parameters['file path to reference meas.'].value.strip():
+            self.apply_reference_to_data(data=data)
+        else:
+            data['values']['referenced transmission [dB]'] = []
+
+        # if user wants to save disk space by discarding raw data, do it now
+        if parameters['discard raw transmission data'].value:
+            data['values']['transmission [dBm]'] = []
```
marcoep, line 218: @delmedico @boris-vukovic 

Check if you like the new implementation. I added another boolean parameter for the user to selectively discard the raw measurement data.

To note is that all save files now have the three data vectors (`wavelength [nm]`, `transmission [dBm]`, `referenced transmission [dB]`) always present as keys in the `data['values']` dict. If you did not chose to reference the data, the data is an empty list, similarly if you chose to discard the raw data.

Check if you like that solution and I'd be happy for some feedback.

marcoep:
@delmedico @boris-vukovic pls review

marcoep:
@delmedico @boris-vukovic pls review

LabExT/Measurements/InsertionLossSweep.py
delmedico: It's fine but I am not fully convinced by this approach. For one because now you always have to check the data arrays before reading some data and then based on that decide which one to further use (in post processing). And two, now in the main window plot you cannot simultaneously show calibrated and uncalibrated data. It would still be more natural to me if it is the calibrated data that is stored in `transmission [dB]` .

LabExT/Measurements/InsertionLossSweep.py
```diff
@@ -181,7 +207,63 @@ def algorithm(self, device, data, instruments, parameters):
...
+        try:
+            with open(ref_fp, 'r') as fp:
+                ref_raw_data = json.load(fp=fp)
+        except FileNotFoundError:
+            raise FileNotFoundError(f'Reference file "{ref_fp:s}" not found.')
+        except json.decoder.JSONDecodeError:
+            raise RuntimeError(f'Reference file "{ref_fp:s}" could not be JSON decoded.')
+
+        # check reference parameters
+        if ref_raw_data['measurement name'] != self.name:
```
delmedico, line 238: Do you think this is strictly necessary? What about only checking that the parameters are present in the measurement settings (like you did below) and also checking for the correct keys in the `data['values']` dictionary? Then we could still apply this in measurements that are based on InsertionLossSweep but with additional features (like applying or optimizing for voltage before performing the sweep)

LabExT/Measurements/InsertionLossSweep.py
marcoep: So, we have discussed this in the last LabExT meeting. We unanimously agree that it is a terrible idea to have two traces with different meanings saved in the same vector name. You will have to adapt your analysis scripts.

And yes, the current implementation does not allow showing two differently named data vectors at the same time in the same plot axis. You can still open the extra plots window (double-click on measurement data plot) and view the two plots side-by-side.

LabExT/Measurements/InsertionLossSweep.py
marcoep: Thank you for this valuable suggestion. Removed the check in [f2d54f0](https://github.com/LabExT/LabExT/pull/168/commits/f2d54f081abd63d36820fc78c988dd58af716940).

marcoep:
@boris-vukovic pls approve s.t. this thing can be merged to dev

LabExT/Measurements/InsertionLossSweep.py
```diff
@@ -181,7 +207,61 @@ def algorithm(self, device, data, instruments, parameters):
...
+
+        assert self.ref_data is not None, 'Reference data is not loaded!'
+
+        # get data back to numpy arrays
+        rec_wl = np.array(data['values']['wavelength [nm]'])
+        rec_tm = np.array(data['values']['transmission [dBm]'])
+
+        # notify user if wavelengths differ > 1pm
+        if any(np.abs(rec_wl - self.ref_data['wl']) > 1e-3):
+            self.logger.warning('Referenced and recorded wavelengths differ by > 1pm!')
```
boris-vukovic, line 260: If one uses a smaller wavelength span than the calibration, this will raise an Exception. This came to mind because our maximal power of our lasers drops the further away from 1550 nm we are and someone might want to measure are higher power. Then he would need to perform a new calibration.

boris-vukovic (approved):

PR #169, opened 2023-07-25
Allow changes to chip name if no chip imported

delmedico:
This is a super small change. Basically it allows the user to change the name of the chip in the main window after a measurement is executed. Meaning, one can perform two measurements with different chip names in the same LabExT session without closing and reopening the GUI. 
Only if a chip is imported, the chip name cannot be changed anymore. This holds also true if a chip is loaded from the last session.


marcoep:
Great UX improvement - thank you!

marcoep (approved):

PR #170, opened 2023-08-03
live viewer code and performance improvement

marcoep:
Performance issues with the Live Plot have largely been resolved by using polled mode. Also, upgraded the power meter card to enable performant multi-channel polling and displaying.

ToDos:
* [x] write separate, more performance live plot class that uses matplotlib animate API for continuous redrawing
* [x] remove color picker, assign fresh color to each new graph
* [x] implement de-/serializing again for LiveViewer
* [x] separate out functionality of bloated PlotControl: move live-viewer relevant functionality (incl. polling) to separate class
* [x] add annotations feature, s.t. new value is readable on screen (maybe in plot, maybe as separate value label?)
* [x] add 2 buttons in LiveViewer for stopping/starting animation, also stop running cards?
* [x] investigate bug: labext does not seem to finish after closing main window if LiveViewer has been opened.
* [x] add low-pass filter feature to bar plots (optionally enabled)
* [x] see Dillon's bugfix below

Ideas - consider parallelizing with asyncio

See
* https://stackoverflow.com/questions/9786102/how-do-i-parallelize-a-simple-python-loop
* https://stackoverflow.com/questions/73334557/how-do-you-get-tkinter-to-work-with-asyncio

Possibly debug with py-spy and speedscope


marcoep:
@delmedico @boris-vukovic @DJAntivenom It would be great if you could give this MR a little test before our meeting on Tuesday.

I'm functionally done, now just working on getting the unittests working.

DJAntivenom:
In my opinion it would make sense to merge this pull request first. Afterwards I can fix the inevitably resulting conflicts with #173 and then it can be merged once it has been reviewed.

delmedico:
I have tested this only with the GUI without looking into the code, so my feedback only comes from the user experience.
- The instrument card used in the last session are currently not saved. Everytime the live viewer is opened the user has to redefine the instruments to be used. Could they be saved as they are done on the current version?
- For me at least, loading new instrument cards (laser and PM) was quite laggy. I doubled check with the current version of the live viewer where loading of new cards is much quicker.
- Colors are now always changing, and the user has no control over it. I would prefer it to either be able to choose the color yourself or always stick to the same scheme (i.e. first trace is standard color blue, second one orange etc.)
- Would it be possible to make the level bar on the right of the plot optional?

DJAntivenom:
- [ ] In `LabExT.View.LiveViewer.LiveViewerController.restore_lv_from_saved_parameters` we should check, if the loaded list has length 0

LabExT/Experiments/StandardExperiment.py
```diff
@@ -115,6 +114,29 @@ def read_parameters_to_variables(self):
...
+        if instr_active_in_lv:
+            if 'no' == messagebox.askquestion('Active LiveViewer Instruments Found!', 'Some ToDos will interact ' +  
+                                          'with instruments that are currently active in the live viewer. Concurrent instrument ' + 
+                                          'access might not be supported for these instrument classes. It is recommended to ' +
+                                          'stop all live-viewer cards before proceeding. Are you sure you want to proceed?',
+                                          default='no',
+                                          icon='warning'):
+                return False
+        return True
+
```
marcoep, line 139: Feature added in this PR: when clicking "Run", there is a message popping up if you forgot to pause the live viewer.

The implementation checks if there is a live viewer card active and occupying an instrument that will be used during the measurements. Only then, the warning is shown.

LabExT/View/MainWindow/MainWindowView.py
```diff
@@ -593,18 +592,10 @@ def set_up_control_frame(self):
...
         self.selec_plot.columnconfigure(0, weight=1)
 
-        self.live_plot = PlotControl(self, add_toolbar=True, figsize=(5, 5), polling_time_s=0.5)
-        self.live_plot.title = "Live Plot"
-        self.live_plot.show_grid = True
-        self.live_plot.data_source = self.model.live_plot_data
-        self.live_plot.grid(row=0, column=2, rowspan=2, padx=10, pady=10, sticky="nswe")
-        self.live_plot.rowconfigure(0, weight=1)
-        self.live_plot.columnconfigure(0, weight=1)
-
```
marcoep, line 607: Main window change: there is no more live plot, only the "saved measurements" plot which occupies now most area of the main window.

marcoep:
Closing this PR in favor of four smaller ones:
* #176 
* #177 
* #178 
* #179

PR #171, opened 2023-09-06
Console logging

DJAntivenom:
This feature allows the user to specify what should be logged to the console. The channels used for logging are now seperated: debug/info -> stdout, warning/error -> stderr.

### DIFF
--- a/LabExT/Logs/MaxLevelFilter.py
+++ b/LabExT/Logs/MaxLevelFilter.py
@@ -1,0 +1,33 @@
+
+import logging
+from logging import LogRecord
+
+class MaxLevelFilter(logging.Filter):
+    """
+    This filter allows only the logging of messages up to a certain
+    logging level.
+    
+    Attributes
+    ----------
+    max_level : int
+        The highest logging level that is allowed
+    """
+
+    def __init__(self, max_level: int, name: str = "") -> None:
+        """
+        Initialize a filter that logs only messages with level less than 
+        or equal to max_level.
+        Initialize with the name of the logger which, together with its
+        children, will have its events allowed through the filter. If no
+        name is specified, allow every event.
+
+        Parameters
+        ----------
+        max_level : int
+            The highest logging level (included) to allow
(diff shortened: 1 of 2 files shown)


DJAntivenom:
@boris-vukovic Could you have a look at this feature I created?

boris-vukovic (approved):

marcoep:
Hi, quick question which I do not get from your description: is there any change in the default behavior?

I would want to keep errors in the LabExT-built-in console!

DJAntivenom:
> Hi, quick question which I do not get from your description: is there any change in the default behavior?
> 
> I would want to keep errors in the LabExT-built-in console!

@marcoep No, the default behaviour is the same. The only thing changing is that the user starting LabExT can now have debug statements in the console that is used to start LabExT. The built-in console stays the same because I use a new `logging.Handler` for the console output (see lines 95-100 in `Main.py`). The built-in console still uses `LabExT.Logs.LoggingWidgetHandler.LoggingWidgetHandler`.

If LabExT isn't started from a console, but through a launcher like the windows start menu (which I think most people do), this whole commit doesn't change anything at all, because then the output streams are dropped anyways. This is more for debugging and for me :)

marcoep:
@DJAntivenom thank you for the clarification! :)

DJAntivenom:
@marcoep One more question, what docstring format should I follow? So far I use the `numpydoc` format, which you can see on lines 6-14 in `LabExT.Logs.MaxLevelFilter.py`.
Does our docgeneration tool support this format or should I use a different one?

marcoep:
@DJAntivenom the LabExT internal code is generally not run through the documentation generation process. Only a little subset (notably the important Instrument / Measurement classes) is actually used to generate the documentation on labext.readthedocs.io.

For how to format the docstrings to be (in theory) automatically generation compatible, see this file: https://github.com/LabExT/LabExT/blob/main/LabExT/Measurements/MeasAPI/Measurement.py

PR #172, opened 2023-09-21
Measurement Wizard Rework

DJAntivenom:
# Summary

This pull request refactors the code for the edit measurement wizard. Instead of using hardcoded functions for each stage/entry in the wizard, we now have an abstract base class for the wizard entries which can be inherited. It provides most of the functionality. The subclasses only need to specify their content and what happens on serializing and deserializing.

Each of the 8 commits integrates one entry in the wizard, except for the first two, which create the new method interface and adapt the `EditMeasurementWizardController` to it.

The methods for the old way of creating gui entries are still in the code and (mostly) unchanged, however, they now log a deprecation warning. They will be removed in the next version of LabExT. Theoretically, no user plugins should ever interact with these methods, but you never know. That's why the methods aren't removed already.

# Advantages

This rework allows us to more easily add new entries to the wizard in the future. There is a small cross dependency between the `EditMeasurementWizardController` and the baseclasses, which could be resolved by using an event-based system. This would then allow the created abstract baseclasses to be used in any wizard. 
Another advantage is that the code is a lot more readable now :)

# Testing

Because the method interface changed I had to adapt one of the test cases to use the new variable names during testing. Afterwards all tests passed. 

## Future ToDo

- Create event-based system to make these classes independently useable
- Rework other wizards in the code to use this method interface

### DIFF
--- a/LabExT/Experiments/StandardExperiment.py
+++ b/LabExT/Experiments/StandardExperiment.py
@@ -358,7 +358,7 @@ class StandardExperiment:
         self.measurements_hashes.remove(mh)
         self.measurements.remove(meas_dict)
 
-    def create_measurement_object(self, class_name):
+    def create_measurement_object(self, class_name) -> Measurement:
         """Import, load and initialise measurement.
 
         Parameters
--- a/LabExT/Measurements/MeasAPI/Measurement.py
+++ b/LabExT/Measurements/MeasAPI/Measurement.py
@@ -7,6 +7,8 @@ This program is free software and comes with ABSOLUTELY NO WARRANTY; for details
 
 import logging
 
+from typing import List
+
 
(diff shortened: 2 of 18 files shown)


DJAntivenom:
The tests fail because there is no file `instruments.config` in the testing environment. Do you know where this file comes from? @boris-vukovic

DJAntivenom:
@boris-vukovic My mistake, I patched a wrong method in the test case.

marcoep:
@DJAntivenom Well done. Great improvement to the code compared to what I've written 3 years ago :D

boris-vukovic (approved):
@DJAntivenom Hey Dillon, looks really good. I also tested it manually with some bad intentions and it works. Thanks!

PR #175, opened 2023-11-20
Repeat more than just the last measurement

marcoep:
Added a small feature to repeat the last N measurements, instead of just the last measurement. Helpful if multiple measurements need to be repeated with similar settings.

When hitting "repeat measurement" button, this dialog will pop up:
image


delmedico (approved):
cool addition

PR #177, opened 2023-11-27
LVI [2/4]: Depreciating "json_able" functions in Widgets

marcoep:
# Goal
Finalize the transition to the new (de-) serialization of parameter and instrument widgets.

# Reasoning
The change on how widgets store their state in JSON in files was started in #172. This PR deletes the former function `make_json_able` and switches over to usage of the new functions.

# Implementation
Some old code in `MovementWizard` needed to be updated. The polygon parameter tables used the old function.

# Compatibility
No compatibility break expected.


DJAntivenom (approved):
Looks very clean

PR #181, opened 2023-11-29
LVI [4/4]: Small Improvements after new LV rewrite: PlotControl performance, Extra-confirmation for run, black config

marcoep:
# Goal
Some clean-up work after the big live-viewer rewrite is in order

# Approach
* **PlotControl performance**: small improvements to the `PlotControl` class should enhance performance: 1st the canvas is only redrawn at the end of `plotdata_changed`, not for every plotted line. 2nd extra code takes care of shutting down the polled execution of foreign functions properly when the plot is destroyed. 3rd polling time for executing foreign functions is increased to 50ms, could possibly be increased to 100ms without perceivable difference.
* **Live-viewer not stopped warning**: A messagebox is now shown to the user in case a measurement run is started but the live-viewer is not stopped. This should avert quite some crashes when using the software.
* **Black formatter** The line-length configuration option for the black code formatter was set to 119 characters.

# Compatibility
No compatibility issues expected.

**--> resubmit due to rebasing error <--**

### DIFF
--- a/LabExT/ExperimentManager.py
+++ b/LabExT/ExperimentManager.py
@@ -12,6 +12,8 @@ import webbrowser
 from os.path import isfile, dirname, join
 from threading import Thread
 
+from typing import TYPE_CHECKING
+
 import matplotlib
 
 from LabExT.DocumentationEngine.Engine import DocumentationEngine
@@ -25,6 +27,11 @@ from LabExT.View.LiveViewer.LiveViewerController import LiveViewerController
 from LabExT.View.MainWindow.MainWindowController import MainWindowController
 from LabExT.View.ProgressBar.ProgressBar import ProgressBar
 from LabExT.Wafer.Chip import Chip
+
+if TYPE_CHECKING:
+    from LabExT.View.LiveViewer.LiveViewerModel import LiveViewerModel
+else:
+    LiveViewerModel = None
 
 matplotlib.use('TkAgg')
 RESOURCE_MANAGER = None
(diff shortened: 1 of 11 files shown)


LabExT/ExperimentManager.py
```diff
@@ -69,6 +76,7 @@ def __init__(self, root, log_file_path, chip=None, skip_setup=False):
         self.mover = MoverNew(experiment_manager=self, chip=chip)
         self.peak_searcher = PeakSearcher(
             None, self, mover=self.mover, parent=self.root)
+        self.live_viewer_model: LiveViewerModel = None
```
DJAntivenom, line 79: I added some type hinting here. I'm sorry, it's like an addiction for me :sweat_smile:

LabExT/Experiments/StandardExperiment.py
```diff
@@ -114,6 +114,29 @@ def read_parameters_to_variables(self):
         self.param_output_path = str(self.save_parameters["Raw output path"].value)
         makedirs(self.param_output_path, exist_ok=True)
 
+    def ask_user_to_continue_even_if_live_viewer_active(self):
+        """ check if any of the instrument addresses in the ToDo queue are currently active in live viewer """
```
DJAntivenom, line 118: Here in the docstring we could add the return values of the function, but I don't think it's very important.

DJAntivenom (approved):
Thank you for these upgrades. It looks great :) 

(I, as a nerd, also like the way you structured the PRs ^^)

PR #182, opened 2023-11-30
Add push-button referencing to live viewer

marcoep:
## Goal
Directly see insertion loss in live viewer by comparing to a set reference.

## Approach
There are now three extra buttons in the live viewer for setting/clearing/recalling reference values. These operate on all visible traces in parallel.

## Implementation
The reference values are currently stored/recalled based on the displayed trace label. Using this to associate a reference value might not be optimum, better would be to use the 3-tuple (instrument type, visa address, channel number) as a key for a reference value. But the current implementation gets us to a working referencing system without too much code. Furthermore, it would be great if we could set/reset/recall references per displayed trace and not operate on all traces in parallel.

## Compatibility
No compatibility breaks.

## Live Viewer GUI
note the three buttons w.r.t. referencing on the bottom of the control frame
image

### DIFF
--- a/LabExT/View/LiveViewer/LiveViewerController.py
+++ b/LabExT/View/LiveViewer/LiveViewerController.py
@@ -145,10 +145,34 @@ class LiveViewerController:
     def reference_set(self):
         for plot_trace in self.model.traces_to_plot.values():
             plot_trace.reference_set()
+        # dump reference values to file for later recall
+        reference_data = {plot_trace.line_handle.get_label(): plot_trace.reference_value for plot_trace in self.model.traces_to_plot.values()}
+        fname = get_configuration_file_path(self.model.references_file_name)
+        with open(fname, "w") as json_file:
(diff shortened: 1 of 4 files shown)


boris-vukovic:
Cool feature! Maybe something to optimize: It would be nice if the reference value do not get deleted if Stop PM and Start PM are pressed

marcoep:
> Cool feature! Maybe something to optimize: It would be nice if the reference value do not get deleted if Stop PM and Start PM are pressed

Well spotted, @boris-vukovic! This was indeed a bug - references should always be recalled even after re-starting power meter cards. Fixed in [4ce500d](https://github.com/LabExT/LabExT/pull/182/commits/4ce500da4ed35b1ef630e1200e510bb9db4fa4ca).

marcoep:
@boris-vukovic and @DJAntivenom - removed emojis!

PR #183, opened 2023-12-04
Use blitting in live plot to improve performance

marcoep:
## Goal
Improve performance of live viewer by using selective redrawing.

## Approach
Matplotlib's animation package offers the functionality of blitting - i.e. only redrawing the changed artists. We make use of this feature which requires tracking of the changed artists each animation tick.

I get between ~15FPS on my MacBook Pro for 2 enabled very noisy power meter signals.

## Implementation
* The default `axis.legend()` call is not compatible with blitted animations. So therefore I simply create my own small "legend" by using `axis.annotate()` calls.
* For debugging performance, I included an FPS counter in the live plot (disabled by default, can be enabled with global setting).
* setting x-axis and y-axis limits is very heavy on performance - so I check if its necessary (i.e. last limits within certain bounds of new limits) and only update then.

## Compatibility
No compatibility breaks.

## Live Viewer GUI
FPS counter and new legend is marked.
image

### DIFF
--- a/LabExT/View/LiveViewer/LiveViewerController.py
+++ b/LabExT/View/LiveViewer/LiveViewerController.py
@@ -97,6 +97,9 @@ class LiveViewerController:
         # averaging for bar plot
         self.model.averaging_bar_plot = max(1, int(abs(parameters["averaging in bar plot"].value)))
 
+        # show FPS counter
+        self.model.show_fps_counter = parameters["show FPS counter"].value
+
     def remove_card(self, card: CardFrame):
         """Removes a card from the liveviewer. This should be called when the user presses the 'x' symbol in the
         top right of a card.
--- a/LabExT/View/LiveViewer/LiveViewerController.py
+++ b/LabExT/View/LiveViewer/LiveViewerController.py
@@ -91,11 +91,8 @@ class LiveViewerController:
         # the minimum y span
         self.model.min_y_span = abs(parameters["minimum y-axis span"].value)
 
-        # enable or disable the bar plots on the right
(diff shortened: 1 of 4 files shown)


DJAntivenom (changes requested):
This is a cool update, however, it doesn't seem to work for me :sweat_smile: 

There are three things:
1. After opening the live viewer and adding a `PowerMeterSimulator` I click 'Start PM'. Then nothing happens and nothing is plotted, but the GUI is still responsive and I have 17 FPS on the empty plotting area. If I then click on 'Update Settings' or 'Stop PM' the whole of LabExT freezes and is completely unresponsive for about 20 seconds. 
2. If I add a `LaserSimulator` and start it, the other buttons work fine and nothing freezes, however, nothing happens on the plot. It just remains empty.
3. In `LabExT.View.LiveViewer.LiveViewerView.ControlFrame.__init__` you use emojis. (This was added in a previous commit, but I must have missed it.) This is a cool idea, however, it is quite dangerous. On my Windows 10 PC I was missing a font and I kid you not, `tkinter` nuked my entire display driver :joy:. It really wasn't built around these kinds of ascii-codes. After installing the required font it worked, but finding that out took quite some time, which I think we shouldn't put our users through XD

boris-vukovic:
Cool feature, it worked for me. However, I probabely had the emoji font installed. As @DJAntivenom suggested, I would remove it for sake of compatibility.

DJAntivenom (approved):

marcoep:
> Cool feature, it worked for me. However, I probabely had the emoji font installed. As @DJAntivenom suggested, I would remove it for sake of compatibility.

Solved in #182 .

PR #184, opened 2023-12-07
Colored logging output

DJAntivenom:
The output in the loggin window (and in the terminal) is now colored.
Screenshot from 2023-12-07 16-44-02


marcoep (approved):
great!

PR #186, opened 2024-02-03
Bugfix and small refactor

delmedico:
- In `Utils.py` there is some small refactoring like type hinting and replacing a deprecated package.
- bugfix:  in `DeviceSelect.py` do not convert device id into `int` as it should be a string and can contain any ASCII character
- deleted `transformations.py` which contained legacy transformations for the old mover and is not used anymore


boris-vukovic (approved):

DJAntivenom (approved):
Great fixes :)

marcoep (approved):
that `importlib_metadata` is a cool find! :D

PR #189, opened 2024-02-26
Measurement Control Settings

delmedico:
This PR adds measurement control settings to LabExT (accessible from the settings menu in the MainWindow) to set:

- the number of finished measurement that are plotted and stored 
- the maximum number of ToDo items that are displayed
- if json files should be prettified (indented)

All of the settings can have a significant impact on the performance of LabExT especially when a large number of measurements are planned and executed.

Measurement control settings window accessed from the MenuBar:
grafik

Example of limited displayed ToDo queue with maximum number of 10:
grafik


LabExT/Experiments/StandardExperiment.py
```diff
@@ -457,6 +462,11 @@ def load_measurement_dataset(self, meas_dict, file_path, force_gui_update=True):
         # add file path to dictionary
         meas_dict["file_path_known"] = file_path
 
+        if self._meas_control_settings.finished_meas_limited:
+            while len(self.measurements_hashes) >= self._meas_control_settings.max_finished_meas:
+                self.measurements_hashes.pop(0)
+                self.measurements.pop(0)
```
marcoep, line 468: I recommend improving the logic here a bit:

Search for the oldest measurement using the `timestamp_known` key. Then, find the hash of that measurement and delete that hash from both of the lists you modify here. I think this would make most sense here - and matches most what you wrote in the tooltip for the corresponding setting.

LabExT/View/MeasurementControlSettings.py
```diff
@@ -0,0 +1,220 @@
...
+import logging
+import os.path
+from tkinter import Toplevel, Checkbutton, BooleanVar, Label, StringVar, Entry, Button
+from tktooltip import ToolTip
+
+from LabExT.Utils import get_configuration_file_path
+from LabExT.View.Controls.CustomFrame import CustomFrame
+
+
+class MeasurementControlSettings:
```
marcoep, line 17: very cool concept of having this "settings" class with a store/load method built in!

LabExT/View/MeasurementControlSettings.py
```diff
@@ -0,0 +1,220 @@
...
+    def save_and_close(self) -> None:
+        self._validate_entries()
+        self._settings.finished_meas_limited = self.measurements_limited.get()
+        self._settings.max_finished_meas = int(self.measurement_limit.get())
+        self._settings.displayed_todo_limited = self.todos_limited.get()
+        self._settings.max_displayed_todo = int(self.todo_limit.get())
+        self._settings.json_indented = not self.no_json_indentation.get()
+
+        self._settings.save_to_file()
+        self.window.destroy()
```
marcoep, line 126: I suggest some functionality here to ask the main window to redraw the todo table here to immediately reflect any changes in the settings that the user made.

I'm pretty sure we used the "reload todo tables" functionality exists somewhere and can be reused.

LabExT/View/ToDoTable.py
```diff
@@ -115,13 +117,20 @@ def regenerate(self):
...
         leftover_hashes = list(self._tree.get_children(item=""))
 
+        n = len(self._original_todo_list)
+        max_displayed_half = self._meas_control_settings.max_displayed_todo // 2
         # compare SOLL vs IST list of what we need to display
         for tidx, todo in enumerate(self._original_todo_list):
 
+            if self._meas_control_settings.displayed_todo_limited:
+                if max_displayed_half - 1 < tidx < n - max_displayed_half:
+                    continue
```
marcoep, line 132: Too simple implementation.

Non-plus-ultra would be a scrollable selection of which ToDo's to display.

In the minimum though, I'd display a row of just `...` to show the user that there are more (not-displayed) Todos.

marcoep (changes requested):

LabExT/Experiments/StandardExperiment.py
delmedico: I have changed this now according to your suggestions. Before it was FIFO (first in first out) based on when it was added to the measurement list `self.measurements`, whereas now it's FIFO based on the timestamp of each loaded measurement. This means that if you add/load an older measurement it will be the first one to be removed again if the limit is reached and even if it was the last measurement that was added.

LabExT/View/MeasurementControlSettings.py
delmedico: good call! We now call the `update_tables()` of the `MainWindowController` on the save_and_close call. This will update the ToDo table based on the new settings immediately. For the measurement table however, this does currently not work, as the adding and removing of measurements is handled in the `StandardExperiment`. For this to work, we would need to separate the loaded meausrements from the `StandardExperiment` and this would require a bit more work.

LabExT/View/ToDoTable.py
delmedico: Thank you for this suggestion. I added one row of `...` to replace all Todos which are not displayed.

marcoep (changes requested):
Code looks good to me.

However, please fix failing tests and submit re-review.

marcoep (approved):
Good to go!

PR #190, opened 2024-03-01
Allow adding Chip sources as Addons

marcoep:
# Goal
Enable easy loading of LabExT addons for additional chip manifest sources.

# Approach
So far, LabExT only supported two simple manifests and offered hard-coded manifest importers for these: a CSV-style Phoenix Photonics one and a custom format by IBM in json-style. A Wizard-style importer was added that can be extended with the usual LabExT-addons-classes that allows to custom-code any kind of importers, that in the end need to return an instantiated `Chip`.

# Implementation
This functionality makes use of the generic `Wizard` GUI element. The addons which are loaded for the manifest importer functionality must subclass `ChipSourceStep` - which is in turn a step of the wizard. This means that the addon class must provide a `build` method, which fills a frame with some GUI elements (e.g. choosing the right file). Eventually, the addon class must call `submit_chip_info` with some list of devices to signal completion of import. The previous functionality of offering to reload the same chip on LabExT restart is retained.

# Compatibility
It did break compatibility with some previous tests - all of which have been fixed. The `Device` class had some constructor parameters' order changed and all (except extra params) are  now mandatory. The `Chip` class also now has all constructor params mandatory. This potentially breaks custom versions of LabExT.

# Future Work
An immediate extension of this work is extending the existing data export functionality to complete the full measurement dataflow (manifest in, measurement execution, manifest-associated-data out). An open design decision is to define separate Addon classes for the exporter or re-use the here provided importer ones.

# Screenshots

#### 1st step - choosing which importer class to use
image

#### 2nd step - configuring importer
image

#### 3rd step - the wizard shows an overview over all imported devices
image


delmedico:
hey, I get an error creating a new measurement and selecting a device from a successfully imported chip.

`
LabExT/LabExT/View/EditMeasurementWizard/WizardEntry/DeviceSelect.py", line 64, in deserialize
    dev_id = int(settings['dev_id'])
ValueError: invalid literal for int() with base 10: '0101A00'
`

This is already corrected in the latest dev version, I think. Maybe it makes sense at this point to change some of the device_id to strings (containing characters) in our example manifests for testing to cover anything similar in future changes.

marcoep:
@delmedico added your suggestion - pls review

delmedico (approved):
This is super nice. Thank you! Additionally, I tested everything with my own `ChipSourceStep` loaded with addons and custom manifest (also json file). Everything worked well!

PR #191, opened 2024-03-04
Replacing flaky with pytes-rerunfailures

marcoep:
Flaky is a test plugin that we used to re-run a failed test up to three times. This was necessary as our GUI integration tests sometimes did not have Tk not properly initiated.

However, [Flaky seems to be an abandoned project](https://github.com/box/flaky/issues/192).  Furthermore, all Python 3.10 tests started to fail on 4th March 2024 due to an import error within the Flaky module.

This PR replaces flaky with the pytest internal [pytest-rerunfailures](https://github.com/pytest-dev/pytest-rerunfailures). This should keep us future-proof in our testing.


delmedico (approved):
cool, it worked for me locally with python 3.10

PR #194, opened 2024-03-14
Save chip info in MainWindow more often than just on exit

marcoep:
# Goal
Make LabExT more user friendly by saving entered chip params more often than just on graceful exit. Sometimes, LabExT crashes (gasp!), and then the chip parameters need to be re-entered, because they weren't saved because there was no graceful exit!

# Approach
Any button that is in the main window now also calls saving of the chip parameters to disk. As the chance of pressing any of the buttons is very high once the chip settings were adapted, this serves as a good proxy to "save immediately". This also works when using the keyboard shortcuts!

# Compatibility
No compatibility issues expected.


delmedico (approved):
good idea!

PR #196, opened 2024-03-27
Fix bug causing LabExT to crash when trying to sort table

EthanLowenthal:
In the Addon Settings menu, LabExT would crash if you clicked on the header of any of the tables.

For example,  if you were to click on the "Imported From"  header:

```
+--------------------+--------------------------------------------+
| Measurement Class  |               Imported From                |
+--------------------+--------------------------------------------+
| DummyMeas          |                                            | <- resolved to float("+inf")
| ->                 | /home/.../LabExT/.../DummyMeas.py          | <- resolved to string
| InsertionLossSweep |                                            |
| ->                 | /home/.../LabExT/.../InsertionLossSweep.py |
+--------------------+--------------------------------------------+

```
And python would try to compare a float and a string and LabExT would crash. Now, the sort function makes sure all the values it is sorting are of the same type.

This does leave a little weird behavior. The above table sorts to:
```

+--------------------+--------------------------------------------+
| Measurement Class  |               Imported From                |
+--------------------+--------------------------------------------+
| ->                 | /home/.../LabExT/.../DummyMeas.py          |
| ->                 | /home/.../LabExT/.../InsertionLossSweep.py |
| DummyMeas          |                                            |
| InsertionLossSweep |                                            |
+--------------------+--------------------------------------------+
```

I don't know if this is a problem. Another fix could be to utilize the underlying treeview and make the rows  actual children of their respective parent, but this changes the behavior of `CustomTable`.

### DIFF
--- a/LabExT/View/AddonSettingsDialog.py
+++ b/LabExT/View/AddonSettingsDialog.py
@@ -110,7 +110,8 @@ class AddonSettingsDialog:
         CustomTable(parent=loaded_meas_frame,
                     columns=('Measurement class', 'imported from'),
                     rows=self._get_loaded_measurements_and_paths(),
-                    selectmode='none')  # custom table inserts itself into the parent frame
+                    selectmode='none',
+                    sortable=False)  # custom table inserts itself into the parent frame
 
         #
         # table with paths to all loaded instruments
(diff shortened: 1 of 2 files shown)


marcoep:
Hell @EthanLowenthal! Many thanks for your contribution. I think though that the most thorough solution to the bug you found is to add a parameter to the `CustomTable` constructor to selectively dis-allow sorting of the table.

For the "Addon Settings" dialog, sorting the tables should be disallowed - we never intended this one to be sort-able anyhow.

It would be great if you could add this to this PR.

EthanLowenthal:
Done!

marcoep (approved):
thank you for the bugfix!

PR #197, opened 2024-04-16
Feature/advanced plotting/file preperations [1/6]

DJAntivenom:
# More Modern Plotting

It's time to add more options and settings to the plotting. This feature will be split up in many subfeatures, each one of which having its own pull request.

The goal in the end is to support line and contour plotting for now, with the possibility to easily add more kinds in the future, if needed.

## This PR

With this PR the main class-layout and file structure is created. After this feature there will most likely be a very big restructuring of the whole codebase, so this might change in the future. However, from the pov of this feature the structures are relatively fixed.

## Compatibility

This feature shouldn't break anything except if someone has written some code specifically dependent on the old plotting module.


marcoep (approved):
lgtm

LabExT/View/Controls/Plotting/PlotView.py
```diff
@@ -0,0 +1,119 @@
...
+        self._data_changed_callback = on_data_change
+
+        self._value = 0
+        self._axes = axes
+        self._change_button = tk.Button(self, text="Click me", command=self.on_click)
+        self._change_button.pack(side=tk.TOP, fill=tk.BOTH)
+
+    def on_click(self):
+        self._axes.clear()
+        import numpy as np
```
boris-vukovic, line 113: I would move the import to the top of the file to make the code clearer and to be conform with PEP8.

boris-vukovic (approved):

LabExT/View/Controls/Plotting/PlotView.py
DJAntivenom: Done in #200

PR #199, opened 2024-04-21
Feature/advanced plotting/plottable data handler [3/6]

DJAntivenom:
# Plottable Data Handler

The new class `PlottableDataHandler` is responsible for preparing the correct data-format for the new plotting views and notifying all registered listeners, when this data changes.

The data itself is a read-only wrapper around a standard dictionary containing the measurements selected by the user.

### DIFF
--- a/LabExT/Experiments/StandardExperiment.py
+++ b/LabExT/Experiments/StandardExperiment.py
@@ -233,8 +233,7 @@ class StandardExperiment:
             data["measurement name and id"] = measurement.get_name_with_id()
             data["measurement id long"] = measurement.id.hex
             data["instruments"] = measurement._get_data_from_all_instruments()
-            data["measurement settings"] = {}
-            data["measurement_params"] = {name: param.value for name, param in measurement.parameters.items()}
+            data["measurement settings"] = {name: param.as_dict() for name, param in measurement.parameters.items()}
             data["values"] = OrderedDict()
             data["error"] = {}
 
--- a/LabExT/Experiments/StandardExperiment.py
+++ b/LabExT/Experiments/StandardExperiment.py
@@ -234,6 +234,7 @@ class StandardExperiment:
             data["measurement id long"] = measurement.id.hex
             data["instruments"] = measurement._get_data_from_all_instruments()
             data["measurement settings"] = {}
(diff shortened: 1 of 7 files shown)


LabExT/Experiments/StandardExperiment.py
```diff
@@ -234,6 +234,7 @@ def run(self):
             data["measurement id long"] = measurement.id.hex
             data["instruments"] = measurement._get_data_from_all_instruments()
             data["measurement settings"] = {}
+            data["measurement_params"] = {name: param.value for name, param in measurement.parameters.items()}
```
marcoep, line 237: This information was meant to be stored under the `"measurement setting"` key, which is defined in the line above.

Originally, we mandated the person who writes the `Measurement` implementation to include saving of the measurement parameters into the `data` dict - see for example [the implementation in InsertionLossSweep](https://github.com/LabExT/LabExT/blob/8d4f1fa2fba6980ad149494aef8de12b522ddb6b/LabExT/Measurements/InsertionLossSweep.py#L146). The original reasoning for this was that the measurement code might actually change parameters (e.g. if a laser wavelength was chosen that the connected laser does not support) and then execute with the new parameters.

However, I think its good to re-think this mandate. I'd suggest you implement the measurement parameter copying from InsertionLossSweep here but save the params in the originally intended `"measurement settings"` key. In your plotting code, then use "measurement settings" instead of "measurement_params". Otherwise we have the same information stored twice in the output JSON, which is not nice.

marcoep (changes requested):

boris-vukovic (approved):

LabExT/Experiments/StandardExperiment.py
DJAntivenom: I saw the `"measurement setting"` key, however, I thought this might cause problems later, if some measurement implementations don't use the right structure or similar.

I think you're right though, the way you described is clearer. I will change it :)

DJAntivenom: Done in 0731ff18d3a070c9afb71ce4ab3af912532bb684.

marcoep (approved):

DJAntivenom:
This PR was superseded after rebasing with the new PR #209

PR #204, opened 2024-04-23
Sweeps now also handles lists -> unconditionally cycle through all list entries.

marcoep:
Just adding the option to setup a sweep through all entries in a `MeasParamList` parameter of a measurement.

Also, fixed two bugs:
* `MeasParamInt` was not able to be swept as it tried to assign a float value to an int param
* There was an off-by-one error setting up the sweep of type `step_size`.

image

### DIFF
--- a/LabExT/Measurements/MeasAPI/Measparam.py
+++ b/LabExT/Measurements/MeasAPI/Measparam.py
@@ -46,7 +46,7 @@ class MeasParam:
         return d
 
     @property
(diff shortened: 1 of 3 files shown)


marcoep:
@boris-vukovic pls review - contains important bugfix!

marcoep:
or @delmedico for that matter

boris-vukovic (approved):
Important feature and fix, thanks!

PR #216, opened 2024-05-22
Update wizard for multiple devices/measurements

delmedico:
# Goal
This is an update of the experiment wizard which allows you to initialize multiple devices and measurements at the same time. The main goal here is to neatly pack everything into the new wizard framework to avoid multiple individual window pop-ups.

# Changes

- This is mainly a GUI change where all the content of the `ExperimentWizard` is brought to one single module using the `Wizard` as a base.
- The parameter sweep feature is now added here as well allowing the user to perform a parameter sweep for every measurement / device selected.
- The device selection table has one additional column showing the original index of the row entries after which all selected devices are then ordered compared to before where they were ordered after the device ID. This way, the original order given by the chip manifest is preserved.
- The generation of the device selection list is improved being now able to handle large lists of already marked and unmarked devices. 

# Compatibility
It did break the previous test of the ExperimentWizard which has been fixed. 

# Screenshots
#### Device selection
Bildschirmfoto 2024-05-29 um 12 01 04

#### Measurement selection
Bildschirmfoto 2024-05-29 um 12 02 09

#### Instrument selection
Bildschirmfoto 2024-05-29 um 12 02 19

#### Parameter selection
Bildschirmfoto 2024-05-29 um 12 02 26

#### Parameter Sweep
Bildschirmfoto 2024-05-29 um 12 03 16


delmedico:
Once this has been reviewed, I could also include removing all the old/unused modules of the ExperimentWizard here.

LabExT/Experiments/StandardExperiment.py
```diff
@@ -53,7 +56,9 @@ class StandardExperiment:
     """StandardExperiment implements the routine of performing single or multiple measurements and gathers their
     output data dictionary."""
 
-    def __init__(self, experiment_manager, parent: Tk, chip, mover=None):
+    def __init__(
+        self, experiment_manager: ExperimentManager, parent: Tk, chip: Chip, mover: Union[Type[MoverNew], None] = None
+    ):
```
DJAntivenom, line 61: Very nice :smile:

DJAntivenom (approved):
Looks good :smiley:

LabExT/Tests/View/ExperimentWizard_test.py
```diff
@@ -222,17 +223,17 @@ def test_experiment_wizard(self):
 
         # randomly set InsertionLossSweep properties
         random_ilm_props = {
-            "wavelength start": random.randint(1460, 1550),
-            "wavelength stop": random.randint(1550, 1640),
-            "wavelength step": random.choice([1.0, 2.0, 5.0, 10.0, 20.0, 25.0, 50.0]),
+            "wavelength start": 1540,
+            "wavelength stop": 1560,
+            "wavelength step": 20,
```
marcoep, line 228: haha, why did you take out the random values?

marcoep (commented):
Great PR, thank you! Definitely necessary.

marcoep:
@delmedico - we reviewed this one today, @EthanLowenthal will add (or just did add) functionality for parameter sweeps.

Pls go ahead and remove the old, unused modules and we're good to merge :)

LabExT/Tests/View/ExperimentWizard_test.py
delmedico: I thought I only want deterministic behaviour in my test :) can also reverse it. Plus, I guess this is more about testing the measurement InsertionLossSweep itself which would be better done in a different test.

DJAntivenom: I agree with Nino

Issue #205, opened 2024-04-30
Measurement settings indistinguishable from meta data in sweep association file


boris-vukovic:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior. Should be as few as possible.

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - Operating System: [e.g. Windows]
 - Python environment [e.g. CPython, 3.9]
 - LabExT Version or Git revision [e.g. v2.1.0]

**Additional context**

marcoep:
@DJAntivenom - you will probably want to use a MultiIndex in your pandas dataframe which tracks the different measurements in a sweep. With one category `metadata` containing id, name, filepath etc. and another category `measurement settings` containing the swept paramters.

See here for examples: https://stackoverflow.com/questions/21443963/pandas-multilevel-column-names

PR #217, opened 2024-05-26
Separate metadata in summary file

DJAntivenom:
# Issue

In issue #205 the problem is described. 
TLDR; The metadata and the measurement settings of sweep members were not stored separately.

# Changes

The dataframe used to store the measurement settings and metadata now contains a multi-index to distinguish the two categories, as described by @marcoep's comment in #205. This likely won't break anything, because all the places in the code, where this dataframe was used, were updated.


LabExT/Experiments/StandardExperiment.py
```diff
@@ -242,15 +242,20 @@ def run(self):
...
+                for _, row in sweep_params.iterrows():
                     # iterate through the rows (i.e. Measurements) belonging to this sweep and store
                     # to metadata
-                    data["sweep_information"]["sweep_association"][row["name"]] = {
-                        name: value for name, value in zip(row.index, row)
+                    temp_dict = dict()
+                    temp_dict["metadata"] = {
+                        name: value
+                        for name, value in zip(row["metadata"].index, row["metadata"])
+                        if name not in ["finished", "name"]
```
marcoep, line 252: why do you need the if statement here? What happens if we have an actual metadata called "name"? Let's discuss in tomorrow's meeting.

marcoep (commented):
Thank you :) Let's discuss the details tomorrow.

DJAntivenom:
@EthanLowenthal If you could quickly review the changes (mainly in `SaveButtons.py`) we can merge it into dev.

EthanLowenthal (approved):

PR #220, opened 2024-05-30
Remove code marked as deprecated

DJAntivenom:
# Goals

The code used for the old measurement creation wizard was marked as deprecated and should be removed.

# Compatibility

Honestly, I would be extremely surprised if this broke anyones setup, because it would mean someone wrote some custom code to interact with very internal methods of LabExT.


marcoep (approved):

LabExT/View/MenuListener.py
```diff
@@ -269,9 +269,6 @@ def client_extra_plots(self):
         main_window.extra_plots = ExtraPlots(meas_table, main_window.view.frame)
         self.extra_plots_toplevel = main_window.extra_plots.cur_window
 
-    def client_side_windows(self):
-        raise DeprecationWarning("Open side windows is deprecated. Do not use.")
-
```
delmedico, line 274: Does this mean we can also remove the "Side Window" button under the ToDo list in the main window?
If so, can this be included in this PR?

delmedico (approved):

PR #221, opened 2024-06-08
build: Pin pytest version to prevent potential issues

marcoep:


marcoep:
@EthanLowenthal - would be happy if you could review briefly. This should fix CI tests, as apparently pytest>=8.2.2 results in tests failing?

PR #202, opened 2024-04-21
Feature/advanced plotting/legend [6/6]

DJAntivenom:
# Legend for Line Plotting

The user can now specify legend elements in the settings panel.

PR #225, opened 2024-07-02
Feature/modern plotting

DJAntivenom:
# New Plotting

This pull request is the summary of #197 to #202. They add functionality to make plotting more user-configurable.


marcoep (changes requested):
Hi @DJAntivenom - I tried out your features, looks good but I do have some change requests:
* The selected axes don't stay when I deselect all curves and re-select some of them. This is very annoying having to select which vector to plot against which every time. Please store the selected axes for every measurement type to disk - as was done in the old plotting.
* When commenting on a measurement, the user can add a "plot label". It's currently not possible to use this entry as source for the legend text. This entry (if it's a nonempty string) should be the default label to be shown in the legend. 
* Add the timestamp as a possible legend entry -> otherwise it's hard to exactly figure out which curve is which.

And lastly, something that we need to be aware of but does not need immediate fixing:
* I have a few measurements that were taken as part of a sweep and are > 3MB per file. LabExT crashed on my MacBook when trying to plot a filled contour plot with those.

marcoep:
Also, if you merge in the latest `dev` commits, the test issue should be solved.

PR #228, opened 2024-09-26
SmarAct MCS2 driver for 3D and 6D modules

ErikMaris:
@marcoep

Hi Marco,

Last time we discussed the MCS2 driver, you suggested to share the handle between the different instances of Stage6DSmarActMCS2 via a module-level variable called _OPEN_6D_MCS2_HANDLE. I implemented this--could you please have a look whether I did this correctly?

You also mentioned the following tests; try that:
1. you can instantiate the two LabExT stages (one for ch1-3, one for ch4-6) and connect to both of them at the same time
2. you can individually control them (moving, etc)
3. you can simultaneously control them (i.e. send moving commands to both stages shortly after each other s.t. both stages move simultaneously).

I addressed these points in the Notebook LabExT/Movement/Stages/Stage6DSmarActMCS2_test_functions.ipynb. Luckily, everything works as desired and you can verify this via the printed output.

I am looking forward to hear your feedback!

Best wishes,
Erik

P.S. I mistakenly forked from main, but luckily the branch can still be automatically merged with dev.


LabExT/Movement/Stages/Stage3DSmarActMCS2.py
```diff
@@ -0,0 +1,478 @@
...
+        cls.driver_path_dialog = DriverPathDialog(
+            parent,
+            settings_file_path="mcsc_module_path.txt",
+            title="Stage Driver Settings",
+            label="SmarAct MCSControl driver module path",
+            hint="Specify the directory where the module MCSControl_PythonWrapper is found.\nThis is external software,"
+            "provided by SmarAct GmbH and is available from them. See https://smaract.com.")
+        parent.wait_window(cls.driver_path_dialog)
+
+        return cls.driver_path_dialog.path_has_changed
```
marcoep: As far as I understood, this function is not necessary for the MCS2 since you're importing the smaract MCS2 functions in line 17 - without the need to load an uninstallable driver package from config file. 

Please get rid of this function (both driver classes).

In the parent class `Stage` the property `driver_specifiable` is in any case False, so the load_driver method is never called.

LabExT/Movement/Stages/Stage3DSmarActMCS2.py
```diff
@@ -0,0 +1,478 @@
...
+    """Implementation of a SmarAct stage. Communication with the devices using the driver version 2.
+
+    Attributes
+    ----------
+    handle : int
+        MCS handle
+    """
+
+    driver_loaded = MCS_LOADED
+    driver_path_dialog = None
```
marcoep: Please add property `description`, see https://github.com/ErikMaris/LabExT/blob/a54462467635a9538a2131ab275500459a94b6c4/LabExT/Movement/Stage.py#L63

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
```diff
@@ -0,0 +1,508 @@
...
+    """Implementation of a SmarAct stage. Communication with the devices using the driver version 2.
+
+    Attributes
+    ----------
+    handle : int
+        MCS handle
+    """
+
+    driver_loaded = MCS_LOADED
+    driver_path_dialog = None
```
marcoep: same

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
```diff
@@ -0,0 +1,508 @@
...
+        cls.driver_path_dialog = DriverPathDialog(
+            parent,
+            settings_file_path="mcsc_module_path.txt",
+            title="Stage Driver Settings",
+            label="SmarAct MCSControl driver module path",
+            hint="Specify the directory where the module MCSControl_PythonWrapper is found.\nThis is external software,"
+            "provided by SmarAct GmbH and is available from them. See https://smaract.com.")
+        parent.wait_window(cls.driver_path_dialog)
+
+        return cls.driver_path_dialog.path_has_changed
```
marcoep: same

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
```diff
@@ -0,0 +1,508 @@
...
+                if is_two_modules(locator):
+                    locators[ii] = locator + '_Ch1-3'
+                    locators.append(locator + '_Ch4-6')
+                else:
+                    locators[ii] = []
+            locators = sorted(locators)
+            return locators
+        return []
+
+    class _Channel:
```
marcoep: Please do not duplicate code for the _Channel class between files. Refactor this class into a separate file and use it in both the main driver classes.

LabExT/Movement/Stages/Stage3DSmarActMCS2.py
```diff
@@ -0,0 +1,478 @@
...
+            locators = buffer.split('\n')
+            def is_one_module(locator):
+                d_handle = ctl.Open(locator)
+                result = ctl.GetProperty_i32(d_handle, 0, ctl.Property.NUMBER_OF_BUS_MODULES) == 1
+                ctl.Close(d_handle)
+                return result
+            return [locator for locator in locators if is_one_module(locator)]
+        return []
+
+    class _Channel:
```
marcoep: Please do not duplicate code for the _Channel class between files. Refactor this class into a separate file and use it in both the main driver classes.

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
```diff
@@ -0,0 +1,508 @@
...
+
+    def __init__(self, address):
+        """Constructs all necessary attributes of the Stage6DSmarActMCS2 object.
+        Calls stage super class to complete initialization.
+        """
+        self.handle = None
+        self.channels = {}
+        if "Ch1-3" in address:
+            self.Axis = Axis_Ch123
+            self.open_new_connection = True
```
marcoep, line 75: What point does this property serve?

I believe in this implementation you assume that a user always initializes _Ch1-3 first and the others second? That might be an issue, because Ch4-6 cannot be used independently if you rely on an open connection from another driver instance.

It would be better if you check if there's an open connection available and if yes attach to that, if no open a new one. I believe that you can rely on the garbage collector to close any open connections (or keep track of open connections yourself and close if last driver is GC-ed).

LabExT/Movement/Stages/Stage3DSmarActMCS2_test_functions.ipynb
marcoep, line 1: After final approval of this PR, please remove this test file - this does not belong into the main repo. If you want you can move it to the tests folder.

LabExT/Movement/Stages/Stage6DSmarActMCS2_test_functions.ipynb
marcoep, line 1: After final approval of this PR, please remove this test file - this does not belong into the main repo. If you want you can move it to the tests folder.

marcoep (changes requested):
Really great job demonstrating how you tested the drivers! Well done!

I have largely two requests: share the `_Channel` class between the drivers and explain better why `open_new_connection` is needed in the 6D class?

LabExT/Movement/Stages/Stage3DSmarActMCS2.py
ErikMaris: Done

LabExT/Movement/Stages/Stage3DSmarActMCS2.py
ErikMaris: Done -- also added driver_specifiable = False

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
ErikMaris: Done

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
ErikMaris: The new ChannelMCS2 class is in the folder Movement/Stages. Should I create the folder Movement/Channels for consistency?

LabExT/Movement/Stages/Stage3DSmarActMCS2.py
ErikMaris: See previous

LabExT/Movement/Stages/Stage3DSmarActMCS2_test_functions.ipynb
ErikMaris: Done

LabExT/Movement/Stages/Stage6DSmarActMCS2_test_functions.ipynb
ErikMaris: Done

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
ErikMaris: Correct, we assume that the locators with suffixes _Ch1-3 and _Ch4-6 are always connected in this order. Since the list of locators, generated by calling `find_stage_addresses()` is not sorted, this order is retained.

Do I understand correctly that you suggest to use the attribute `self.connected` to check whether an open connection is already available? Unfortunately, we cannot use it to recognise whether a connection has been opened yet, because it is always `self.connected = False` when `Stage6DSmarActMCS2.py` object is instantiated, independently of the other locators.

The purpose of `open_new_connection` is to differentiate between: 
(1) opening a new connection to a new MCS2 module with two bus modules (and assigning the first bus to the object) 
(2) keeping the current connection (in `_OPEN_6D_MCS2_HANDLE`) but assigning the second bus to the object. 
If there would be two MCS2 modules with two busses each, this would be recognised and an error would be thrown. This scenario is not supported, because we only store one global handle in `_OPEN_6D_MCS2_HANDLE` to one MCS2 module.

Or do you mean something like this—removing `open_new_connection` but also the error that checks for the scenario described before (with 2x MCS2 with 2x bus modules each)? 
```
        if _OPEN_6D_MCS2_HANDLE is None:
            _OPEN_6D_MCS2_HANDLE = self._open_system()
            self.handle = _OPEN_6D_MCS2_HANDLE
        else:
            self.handle = _OPEN_6D_MCS2_HANDLE
```

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
marcoep: nope, all good as is

LabExT/Movement/Stages/Stage6DSmarActMCS2.py
marcoep: If you want to be super correct, you could make `_OPEN_6D_MCS2_HANDLE` a dictionary and store the open handles depending on the rest of their USB address...

But your solution here works as well, with a bit more rigorous error checking. Which is good in any  case.

marcoep (approved):

PR #229, opened 2024-09-27
Measurement stepped sweep

ErikMaris:
Hi @marcoep @delmedico 

I tried to perform a stepped sweep with the N7779C tuneable laser and N7744C power meter using a triggering scheme. This requires two delays: a *stabilisation time* after the wavelength of the laser has been changed and an *averaging time* after the laser has send the trigger to the power meter to take a measurement. However, I cannot find how to set the delay in the laser for the averaging time.
Despite being faster, I am not confident enough that the measurements done using this scheme are correct. Therefore, I decided to trigger the stepped wavelength by the computer. The measurement file is contained in this pull request.

Best,
Erik

### DIFF
--- a/LabExT/Measurements/InsertionLossSteppedSweep.py
+++ b/LabExT/Measurements/InsertionLossSteppedSweep.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+LabExT  Copyright (C) 2024  ETH Zurich and Polariton Technologies AG
+This program is free software and comes with ABSOLUTELY NO WARRANTY; for details see LICENSE file.
+"""
+
+import time
+import numpy as np
+
+from LabExT.Measurements.MeasAPI import *
+
+
+class InsertionLossSteppedSweep(Measurement):
+    """
+    ## InsertionLossSteppedSweep
+
+    This measurement uses a tunable laser source to make a spectral measurement by looping over the output signal
+    wavelengths.  The optical power meter takes a measurement after a given stabilization time. The resulting arrays 
+    of wavelength and detected signal samples provide a spectrum showing the wavelength dependence of the DUT.
+
+    Currently this measurement supports Agilent/Keysight swept lasers (model numbers 816x, N777xC) and triggered power
+    meters (model numbers 816x or N77xx).
+
+    All triggers are sent by the computer and no trigger cables need to be connected
+
+    #### laser parameters
+    * **wavelength start**: starting wavelength of the laser sweep in [nm]
+    * **wavelength stop**: stopping wavelength of the laser sweep in [nm]
+    * **wavelength step**: wavelength step size of the laser sweep in [pm]
+    * **sweep speed**: wavelength sweep speed in [nm/s]
+    * **laser power**: laser instrument output power in [dBm]
+    * **stabilization time**: time between changing the laser wavelength and sampling of the power [s]
+
+    #### power meter parameter
+    * **averaging time**: time over which a power measurement is averaged [s]
+    * **powermeter range**: range of the power meter in [dBm]
+
+    #### user parameter
+    * **users comment**: this string will simply get stored in the saved output data file. Use this at your discretion.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)  # calling parent constructor
+
+        self.name = 'InsertionLossSteppedSweep'
+        self.settings_path = 'InsertionLossSteppedSweep_settings.json'
+        self.instr_laser = None
+        self.instr_pm = None
+
+    @staticmethod
+    def get_default_parameter():
+        return {
+            # lower bound for sweep
+            'wavelength start': MeasParamFloat(value=1530.0, unit='nm'),
+            # upper bound for sweep
+            'wavelength stop': MeasParamFloat(value=1570.0, unit='nm'),
+            # step size
+            'wavelength step': MeasParamFloat(value=10.0, unit='nm'),
+            # laser stabilization time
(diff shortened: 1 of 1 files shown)


LabExT/Measurements/InsertionLossSteppedSweep.py
```diff
@@ -0,0 +1,174 @@
...
+        self.instr_laser.wavelength = start_lambda
+        number_of_points = int((end_lambda - start_lambda) / lambda_step) + 1
+
+        # PM settings
+        self.instr_pm.averagetime = averaging_time
+        self.instr_pm.wavelength = center_wavelength
+        self.instr_pm.range = pm_range
+        self.instr_pm.unit = 'dBm'
+
+        time.sleep(1) # allow laser to move to start wavelength
```
marcoep: Please double check if this timed wait is really necessary. Timed waits should not be used, instead poll the instrument if the desired wavelength is reached. Most often, this can be done with something like `OPC?` but I suspect Keysight offers a specific command for checking if wavelength is stable?

LabExT/Measurements/InsertionLossSteppedSweep.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # STARTET DIE MOTOREN!
+        for idx, wavelength in enumerate(wavelengths):
+            self.logger.info("Taking a power measurement at {} nm.".format(wavelength))
+
+            # set laser wavelength
+            self.instr_laser.wavelength = wavelength
+            
+            # wait for the laser to stabilize and let one averaging time pass before taking a sample
+            time.sleep(stabilization_time + averaging_time)
```
marcoep, line 149: See comment above. You can possibly get rid of the stabilization time param as well.

marcoep (changes requested):
Thank you for your contribution! Please double check if you can't query the laser if wavelength is stable at desired position? This would enhance reliability of the measurement and ensure a correct measurement procedure.

LabExT/Measurements/InsertionLossSteppedSweep.py
ErikMaris: I tested whether this wait is necessary, because I observed--in the past--that the point data point was off without wait. I cannot reproduce this and have therefore removed the wait.

However, I did an observation that I cannot explain. There seems to be an offset in the transmission depending on the wavelength sweep range.

wavelength_stop_sweep_1s-startwait

Regardless of timed waits (e.g., between experiments) this behaviour is consistently observed. Any idea what is going on here?

LabExT/Measurements/InsertionLossSteppedSweep.py
ErikMaris: To establish that the laser needs a stabilisation time, I performed a sweep. Only at a stabilisation time of 0.5 s and longer, the measured transmission remained constant.

stabilization_time_sweep

I tried two things, but both did not work:
(1) check whether the laser is busy:
```
while not bool(self.instr_laser.request('*OPC?')):
	print('Wait')
	time.sleep(0.0001)
```
(2) check whether the wavelength has been reached:
```
while not np.isclose(self.instr_laser.wavelength,wavelength):
	print(self.instr_laser.wavelength)
	time.sleep(0.0001)
```
Here, I noticed that the requested wavelength is such a nice round number that it must be the set wavelength. Also, the N7779C does not have, AFAIK, a sensor to measure the absolute wavelength.

My suggestion is to leave these timed waits in the code, however ugly they are.

marcoep: Hi @ErikMaris,

I've scoured the laser manuals to find a way to query the wavelength-tuning-state and if the laser is stable after requesting a tuning through the API - but I did not find anything.

I was also able to confirm with our N7776C unit that simply querying `:SOUR0:WAV?` simply reports the target wavelength and not the actual wavelength. I then tried to use the `*WAI` command to tell the laser to respond only after wavelength setting - but this also does not work.

In [wavelength_tuning_and_api_response.py.txt](https://github.com/user-attachments/files/19787093/wavelength_tuning_and_api_response.py.txt) I've tested the API response when initiating tuning. It seems like that there is a break in API responses during the tuning of the laser (see lines with a bunch of !!!!). If I assume that the API responds after tuning completed, I can see that my laser model needs ~0.7s for tuning. At least on my model, this seems to be tuning range independent (tuning over 1nm, 10nm and 180nm all cause response delays of ~0.7s in the API.) Interestingly, the 0.7s are close to your 0.5s...

Long story short, I think keeping the stabilization time as a parameter is good. However, you will have to carefully fine-tune this parameter based on your other sweep settings. I'll approve the PR, I think this is a good alpha-version of the measurement!

Thanks a lot for contributing.

marcoep (approved):

LabExT/Measurements/InsertionLossSteppedSweep.py
marcoep: Oh my, I do not know what that could be. Please make sure to carefully baseline / calibrate your setup before doing measurements to take care of such effects.

LabExT/Measurements/InsertionLossSteppedSweep.py
ErikMaris: Thanks for looking into this! Good to see that your tuning time is close to the one I found for our N7779C laser

LabExT/Measurements/InsertionLossSteppedSweep.py
ErikMaris: I discussed this issue with @delmedico yesterday. He suggested to request the input power at each wavelength and use this for the computation of the transmission. I will look into this

ErikMaris: Fixed in PR #234

PR #231, opened 2025-04-21
bugfix: missing definition in case of no SmarAct drivers loaded

marcoep:


marcoep:
@erikmaris pls check if LabExT driver runs with this bugfix :)

EthanLowenthal (approved):

PR #237, opened 2025-07-18
remove warning: available stages logging

delmedico:
This is a super small change to remove the logger warnings of stage driver loading errors. Basically we have this information in the "Configure Mover" window anyway. And it is just annoying seeing this warning every time you run LabExT from the terminal or from wherever :)

I put it as debug logging but we can also change it to info logging so it would be saved to the log file (still a bit unnecessary imo).


marcoep (approved):

marcoep:
@delmedico approved - sorry, was on vacation until yesterday

PR #238, opened 2026-01-06
Adding filtering for device selection table

delmedico:
This PR adds the possibility to filter the device table so that only rows / devices are shown that contain a given search string.

Screenshot 2026-01-06 105816

In the _Multi-Device Multi-Measurement Experiment_ the user first has to choose the devices to perform the measurements on. This can be a tedious process when chip manifests are loaded with a large number of devices of various types. 
The user can now filter the shown devices by selecting a search string that should be contained within the defined column value. In the image above, the table is filtered to only show devices that contain _cutback#3_ in their _Type_ value. The user can then mark the selected devices of the filtered table. This filtering can be done multiple times until all the wanted rows are marked.

What is happening in the back?

- `DataFrameTable` replaces `CustomTable`
The `DataFrameTable` is a UI element that is based on and representation of a pandas DataFrame. This makes it very intuitive and easy to work with and allows features like filtering and other data processing routines.

- `CustomTable` has seen some small refactoring and type hinting.
(Actually this is not used in the end, yet still useful)

- The `Wizard` and `ExperimentWizard` saw small refactorings but mostly type hinting.

- `PlaceholderEntry` is a custom version of the tkinter Entry widget that in addition has a greyed-out placeholder value when nothing is written in the entry field.

### DIFF
--- a/LabExT/Tests/View/ExperimentWizard_test.py
+++ b/LabExT/Tests/View/ExperimentWizard_test.py
@@ -109,34 +109,34 @@ class ExperimentWizardTest(TKinterTestCase):
         # select devices step
         #
         device_step = exp_wizard.step_device_selection
+        dataframe_table = device_step.device_table.device_table
         self.assertIsInstance(device_step, DeviceSelection)
-        all_rows = device_step.device_table.device_table.get_tree().get_children()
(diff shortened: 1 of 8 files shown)


marcoep (approved):
good stuff, neat little feature!

